PR #696, opened 2017-10-30
Update location of gtas.war

tgudmund88:
When doing the install, this is where I found the gtas.war file.


originalname51:
This commit has been pushed to dev and will be in master for the 1.3 release.

Issue #873, opened 2018-08-19
Parsing & Loading | Booking date appears to be wrong


dje04001:
The date we're parsing and loading for booking date appears to be off. I think it should come from the RCI line. Per the spec doc,

RCI+TZ:W9TEND::230513:181348' | Airline locator W9TEND, created 23 May 2013 at 18:13Z

We're pulling this date from somewhere else, likely the time from the last update.

originalname51:
When processing message from the [spec](https://www.iata.org/iata/passenger-data-toolkit/assets/doc_library/04-pnr/PNRGOV%20EDIFACT%20Implementation%20Guide%2013_1.pdf) it appears what is expected to be the booking date is getting parsed as the reservation date, see line 114 PnrGovParser.

See [pr](https://github.com/US-CBP/GTAS/pull/875)

PR #875, opened 2018-08-25
Unit test showing RCI line to be investigated for booking date is pop…

originalname51:
…ulated in reservation date. The Payment made date is populated as booking date.

The following unit test first parse slightly modified message from[ page 76 and 77 of the spec](https://www.iata.org/iata/passenger-data-toolkit/assets/doc_library/04-pnr/PNRGOV%20EDIFACT%20Implementation%20Guide%2013_1.pdf) and then assert the booking date is being set to the first date with code 710 and what [issue 873 ](https://github.com/US-CBP/GTAS/issues/873) believes is the booking date is set to the reservation date.


originalname51:
Updated test to read in files from resources for pnr messages.

PR #878, opened 2018-09-02
Configuration for use with IBM MQ 

LeHombre:
How do  set up GTAS to receive/send API/iAPI/PNR though IBMQ? Is there an implementation guide rough or otherwise?

Jon


LeHombre:
Really would like a admin guide!

dje04001:
Hi Jon! Thanks for your interest. We have a basic guide for building the application from source / some admin tasks, but we're working on the IBMQ client piece now. We have two deployments coming up in November that will use it, and we'll share what we learned.

originalname51:
Either of the following will bridge GTAS to IBMMQ
https://activemq.apache.org/jms-to-jms-bridge
or
https://github.com/US-CBP/MQ-Bridge/tree/IbmMqToActiveMq

Issue #901, opened 2018-10-03
Watchlist Displays Passenger Name instead of Watchlist Name hit


originalname51:
Watchlist  displays passenger name multiple times on hit instead of watchlist names that hit on. See pictures. Generated passenger name was SEMIRA SELASSIE.

![image](https://user-images.githubusercontent.com/10320916/46431325-7f616800-c719-11e8-8d98-d1bdc4cdf333.png)

![image](https://user-images.githubusercontent.com/10320916/46431353-943dfb80-c719-11e8-900a-9ab493327b3e.png)

PR #909, opened 2018-10-15
This commit fixes the issue in #901.

originalname51:
Matching Service : Got rid of un-needed public access modifier.

Matching Service Impl : We were setting watch list item name and dob based on the passenger that was a hit instead of the watchlist item. This resulted in the error seen on the front end. Changed to these values from watchlist item


originalname51:
closing - tsegay fixed

PR #928, opened 2018-10-24
Parser update. This will use the segment terminator as a way to guara…

originalname51:
…ntee

the segment is starting. Because this is using a field before the segment,
a check of the release character is now required.

These changes represent a future enhancement of the parser. 

This Pull Request SHOULD NOT be merged in now, until a milestone is reached.


originalname51:
As found in the above commit the edge case is handled already.

Issue #890, opened 2018-09-21
Fuzzy Watchlist Hit indicator on Flights grid


dje04001:
The current state gives no indication to the user for flights that have fuzzy watchlist hits. I'm thinking it might be a good idea to include fuzzy hits in the regular watchlist counter.

dje04001:
and Passengers grid

originalname51:
Creation of rule hits coupled to rule engine. New service will be required to generate new hit details on non-rule engine hits.

originalname51:
Changing sql on passenger query from 
select
        generatedAlias0,
        generatedAlias1,
        generatedAlias2 
    from
        Passenger as generatedAlias0,
        Flight as generatedAlias1,
        FlightPassenger as generatedAlias3 
    left join
        generatedAlias0.hits as generatedAlias2 with generatedAlias2.flight.id=:flightId 
    where
        (
            generatedAlias3.flightId=1L 
        ) 
        and (
            generatedAlias0.id=generatedAlias3.passengerId 
        ) 
        and (
            generatedAlias1.id=1L 
        ) 
    order by
        generatedAlias2.watchListHitCount desc,
        generatedAlias2.ruleHitCount desc,
        generatedAlias1.eta desc

PR #935, opened 2018-10-30
Issue #890.

originalname51:
Added sort to passenger service impl. Added group by statement to passenger
service impl that returns a distinct passenger.

Updated Dashboard to return distinc results.

Updated flights page to add flag.

Added relationship to pax_watchlist_link table in passenger (lazy)


originalname51:
@tjwdev  reviewed

Issue #837, opened 2018-07-05
query timeouts


dje04001:

originalname51:
Adding two values in admin settings to configure amount of records that can be returned by queries. 

Default value will be 1,000 configurable up to whatever number wanted.

PR #937, opened 2018-11-01
Issue #837

originalname51:
Set max results on queries to stop long running queries.

Set default max results to 1000 for flights and/or passengers.

Set header to show user when query went over 1000 and data may be truncated.

Updated initial scripts to have values.


originalname51:
Updated to remove extra tab.

Issue #880, opened 2018-09-10
Flight History | remove links to "booking detail" flights


dje04001:

originalname51:
There are two types of passenger booking detail flights. Regular flights and booking detail flights.

The purpose of this ticket is to remove the link from booking detail flights and correctly link the passenger details page between flights.

PR #940, opened 2018-11-02
Issue #880

originalname51:
Rewrote some code relating to booking details.

Added a class to return passengerId with flight to fix broken links.


originalname51:
Reviewed by @geoffbays  and @tsegay . Updated to make a single call to the database instead of a looping call. This allowed the streams to be further updated to make them hopefully more readable.

PR #993, opened 2019-02-14
Loader rule runner collisions

originalname51:
This code is ready for test and integration with dev branch.


originalname51:
Changes made allow rule runner and loading and booking detail compression to happen in a near-concurrent way. 
There are still some issues with concurrency related to booking details compression but they should happen quite a bit less frequently and running these 
scheduled tasks at the same time should now not be problematic.

Additional refactoring to break the flight and passenger model up into composite peices, allowing the parent to be 
immutable so the row can be referenced when checking foriegn keys is recommended as a follow up task.

Synopsis of changes:

1. Moved case saving to database OUT of case disposition service. 
    Updated Case creation to use caching. (Hazelcast)
    Refactored/code clean up on case creation.
    Fixed several bugs related to case creation and rule running.
2. Renamed loaderrepository to GtasLoader. Added services. 
Broke apart how objects are loaded, loading flights, then passengers, then other details.
3. Added caching to loader to increase speed on commonly done lookups (hazelcast)
4. Broke apart flight passenger count as updating the flight row is costly and causes deadlocks.
5. Re-wrote booking detail compressor. Enhanced Service Class
6. Re-wrote the rule runner scheduled task to generate java objects and cases 
and then save them all at once. This will reduce liklihood of hanging hit details / hit summaries / cases
7. Deleted and commented out several code parts that are unused.
8. Removed unit test to classes that are unused.
9. Added a message status table so message can be treated as immutable and referenced without deadlocking.

gtas-parent/gtas-commons/src/main/java/gov/gtas/services/BookingDetailServiceImpl.java
```diff
@@ -1,84 +1,147 @@
...
+        oldBookingDetailsPnr.forEach(bd -> {
+            bd.getBookingDetails().add(nbd);
+            bd.getBookingDetails().remove(obd);
+        });
+
+        oldBookingDetailsFlightLeg.forEach(fl ->
+                fl.setBookingDetail(nbd)
+        );
+
+        oldBD.setPassengers(null);
```
originalname51: Do we need to set this as null?

geoffbays:
Rob:
If you are saving or deleting oldBD probably yes. If not, then not.

Geoffrey

Issue #1066, opened 2019-04-16
Passenger Details | APIS tab duplicate flightPax


dje04001:
the query on the APIS tab for finding co travelers is returning a list instead of a set.

Issue #930, opened 2018-10-24
1DL | Encounter


dje04001:
Encountered

In Progress - add user name to a list
    Denied Boarding
    Refused Entry
    Secondary Positive
    Secondary Negative
    Missed

dje04001:
![uploads/86e36bba-9657-4d4a-bdf2-379b8e90e8c6/Picture2.png](https://s3-us-west-2.amazonaws.com/prod.huboard.com/uploads%2F86e36bba-9657-4d4a-bdf2-379b8e90e8c6%2FPicture2.png)

PR #1078, opened 2019-04-29
Dev 930 1066

kasimhelil:
APIS tap: Fixed duplicate records on passenger detail; Solved #1066.
One Day Lookout: Resolved Encounter Status update; Solved #1052 and #930.

### DIFF
--- a/gtas-parent/gtas-commons/src/main/java/gov/gtas/model/Case.java
+++ b/gtas-parent/gtas-commons/src/main/java/gov/gtas/model/Case.java
@@ -6,6 +6,9 @@
 package gov.gtas.model;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
+
+import gov.gtas.enumtype.EncounteredStatusEnum;
+
 import org.apache.commons.lang3.builder.EqualsBuilder;
 import org.apache.commons.lang3.builder.HashCodeBuilder;
 
@@ -101,16 +104,17 @@ public class Case extends BaseEntityAudit {
     @JoinColumn(name = "flightId",insertable=false, updatable=false, referencedColumnName = "id")
     private Flight flight;
     
+    @Enumerated(EnumType.STRING)
     @Column(name = "encountered_status", nullable=false)
-    private String encounteredStatus = "Not Encountered Yet";//The default value for encountered status
+    private EncounteredStatusEnum encounteredStatus = EncounteredStatusEnum.NOT_ENCOUNTERED;//The default value for encountered status
 
  
-	public String getEncounteredStatus() {
+	public EncounteredStatusEnum getEncounteredStatus() {
 		return encounteredStatus;
 	}
 
 
-	public void setEncounteredStatus(String encounteredStatus) {
+	public void setEncounteredStatus(EncounteredStatusEnum encounteredStatus) {
 		this.encounteredStatus = encounteredStatus;
 	}
 
--- a/gtas-parent/gtas-commons/src/main/java/gov/gtas/model/Case.java
+++ b/gtas-parent/gtas-commons/src/main/java/gov/gtas/model/Case.java
@@ -100,8 +100,23 @@ public class Case extends BaseEntityAudit {
     @ManyToOne(fetch = FetchType.EAGER)
     @JoinColumn(name = "flightId",insertable=false, updatable=false, referencedColumnName = "id")
     private Flight flight;
+    
+    @Column(name = "encountered_status", nullable=false)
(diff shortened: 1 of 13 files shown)


gtas-parent/gtas-webapp/src/main/webapp/onedaylookout/oneDayLookout.html
```diff
@@ -1,4 +1,4 @@
-<!-- <div ng-include="'views/sidepanel.html'"></div> -->
+s<!-- <div ng-include="'views/sidepanel.html'"></div> -->
```
originalname51: Is this a stray letter?

gtas-parent/gtas-commons/src/main/java/gov/gtas/model/Case.java
```diff
@@ -100,8 +100,23 @@ public void addHitsDisposition(HitsDisposition _tempHit){
     @ManyToOne(fetch = FetchType.EAGER)
     @JoinColumn(name = "flightId",insertable=false, updatable=false, referencedColumnName = "id")
     private Flight flight;
+    
+    @Column(name = "encountered_status", nullable=false)
+    private String encounteredStatus = "Not Encountered Yet";//The default value for encountered status
```
originalname51: Would an enum value be better used here?

gtas-parent/gtas-commons/src/main/java/gov/gtas/repository/CaseDispositionRepository.java
```diff
@@ -84,6 +84,11 @@ default Case findOne(Long id)
    @Query("update Case set oneDayLookoutFlag = :flag where id = :caseId")
    public Integer updateOneDayLookoutFlag(@Param("caseId")Long caseId, @Param("flag") Boolean flag);
    
+   @Modifying
+   @Transactional
+   @Query("update Case set encounteredStatus = :encStatus where id = :caseId")
+   public Integer updateEncounteredStatus(@Param("caseId")Long caseId, @Param("encStatus")String encStatus);
```
originalname51: Because we will not consume the Integer this would be better as void value.

gtas-parent/gtas-commons/src/main/java/gov/gtas/services/CaseDispositionService.java
```diff
@@ -90,4 +90,6 @@ public Case registerCaseFromRuleService(Long flight_id, Long pax_id, String paxN
     
     public Iterable<RuleCat> findAllRuleCat();
 
+	public Boolean updateEncounteredStatus(Long caseIdAsLong, String encStatus);
```
originalname51: No need for true false value here - this can throw an exception on failure instead.

gtas-parent/gtas-commons/src/main/java/gov/gtas/vo/passenger/OneDayLookoutVo.java
```diff
@@ -17,8 +17,15 @@
 	    private String etaEtdTime;
 	    private String fullFlightNumber;
 	    private String name;
-            private String origDestAirportsStr;
+        private String origDestAirportsStr;
+        private String encounteredStatus;
 	    
+		public String getEncounteredStatus() {
```
originalname51: Consider making this enum

gtas-parent/gtas-webapp/src/main/java/gov/gtas/controller/OneDayLookoutController.java
```diff
@@ -173,5 +173,25 @@ public String getUserLocation(String userId) {
...
 		return result;
 	}
+	
+	@RequestMapping(value = "/encounteredstatus", method = RequestMethod.POST)
+	public @ResponseBody Boolean updateEncounteredStatus(@RequestParam(value = "caseId", required = true) Long caseId,
+			@RequestParam(value = "newStatus", required = true) String newStatus) {
+		boolean result = false;
+		logger.debug("...Encountered status is being updated for case ID: " + caseId);
+
+		if (caseId != null) {
```
originalname51: Remove try/catch and allow exception to be thrown ; front end appears to be written to handle exception.

tjwdev: Test comment, and also don't see try/catch in selected code

originalname51: Sorry; please look at line 184.

PR #1083, opened 2019-05-02
Changed citizenship to Nationality; Resolved #1049

### LINKED ISSUE #1049: Convert Citizenship to Nationality
Citizenship is not a data element in API / PNR messages, and document issuance country is not enough information to infer citizenship.

Citizenship should be removed from Rules and Queries
Citizenship references displayed in the web app should be changed to Nationality.

kasimhelil:


gtas-parent/scripts/elastic/config/kibana/kibana.default-dashboard.json
```diff
@@ -165,9 +165,9 @@
       "updated_at": "2019-02-10T03:22:41.755Z",
       "version": 4,
       "attributes": {
-        "title": "case*",
```
originalname51: This does not appear to be based on nationality.

tsegay: changes in line 170 is related to nationality, I can see `citizenshipcountry` is replaced by `nationality`. L168-L170 and L154-L156 are not related to nationality and it might make sense to put those changes in a separate commit.

Issue #1081, opened 2019-05-01
Watchlist admin | enable all document types in the PNR spec


dje04001:
Right now, users on the front end can only add Passports and Visas to a watchlist (for the individual add function). Please add all document types per the PNR spec

PR #1085, opened 2019-05-06
Dev #1081

kasimhelil:


kasimhelil:
I've tested it with the change @tsegay made and worked well for me!

PR #1187, opened 2019-06-12
Quickmatch update

tsegay:


gtas-parent/gtas-commons/src/main/java/gov/gtas/repository/PaxWatchlistLinkRepository.java
```diff
@@ -7,24 +7,29 @@
...
+	
 	@Query("SELECT p.watchlistItem.id FROM PaxWatchlistLink p WHERE p.passenger.id =:paxId")
 	public List<Long> findWatchlistItemByPassengerId(@Param("paxId") Long paxId);
+	
 	@Modifying
 	@Transactional
 	@Query(value="insert into pax_watchlist_link (last_run_timestamp, percent_match, verified_status, passenger_id, watchlist_item_id) VALUES (:lastDate, :percentMatch, :verifiedStatus, :passengerId, :watchlistItemId)", nativeQuery=true)
 	public void savePaxWatchlistLink(@Param("lastDate") Date lastDate, @Param("percentMatch") float percentMatch, @Param("verifiedStatus") int verifiedStatus, @Param("passengerId") Long passengerId, @Param("watchlistItemId") Long watchlistItemId);
+	
+	@Query(value = "select new map(p.id as passenger_id, i.id as watchlist_item_id) from PaxWatchlistLink pl left join pl.passenger p left join  pl.watchlistItem i")
```
originalname51: I think this information is better used from the getPassengerMatchingInformation. This should save a trip to the database.

gtas-parent/gtas-commons/src/main/java/gov/gtas/services/matcher/MatcherParameters.java
```diff
@@ -1,21 +1,72 @@
...
+		this.dobYearOffset = dobYearOffset;
+	}
+
+	public List<HashMap<String, String>> getDerogList() {
+		return derogList;
+	}
+
+	public void addDerogList(List<HashMap<String, String>> derogList) {
+		if(this.derogList == null)
+			this.derogList = new ArrayList<>();
```
originalname51: From a style point of view I think it would be better to keep  '{' and '}' on if statements as this is more commonly used on the app code.

gtas-parent/gtas-commons/src/main/java/gov/gtas/services/matcher/MatchingService.java
```diff
@@ -5,6 +5,8 @@
 import gov.gtas.model.Flight;
 import gov.gtas.model.MessageStatus;
 import gov.gtas.model.Passenger;
+import gov.gtas.services.matcher.quickmatch.QuickMatcher;
```
originalname51: I think these import statements do not need to be added here.

gtas-parent/gtas-commons/src/main/java/gov/gtas/services/matcher/MatchingServiceImpl.java
```diff
@@ -165,54 +179,90 @@ private MatcherParameters getMatcherParameters(List<Long> passengers) {
...
         return matcherParameters;
     }
 
+	/**
+	 * 
+	 * @return
+	 */
+	@SuppressWarnings("unlikely-arg-type")
+	public Map<Long, Set<Long>> getAllPaxWatchlistLinks() {
+
```
originalname51: A potential replacement to making a db call here would be to pass in the passengers being iterated over and do a for each loop for each passenger by their pax watchlist id. 

The         Set<Passenger> passengers = passengerRepository.getPassengerMatchingInformation(messageIds); line in MatchingServiceImpl brings back a passenger with the pax watchlist hit list already populated.

gtas-parent/gtas-commons/src/main/java/gov/gtas/services/matcher/MatchingServiceImpl.java
```diff
@@ -165,54 +179,90 @@ private MatcherParameters getMatcherParameters(List<Long> passengers) {
...
-                }
-                
-                int dobYearOffset = getDobYearOffset(MatchingContext.DOB_YEAR_OFFSET);
 
-                MatchingResult result = qm.match(passenger, derogList, matcherParameters.getThreshold(), dobYearOffset);
-                
+				Set<Long> wlItemIds = matcherParameters.getPaxWatchlistLinks(passenger.getId());
+				hitCounter += wlItemIds.size();
+				//
+				MatchingResult result = matcherParameters.getQm().match(passenger, matcherParameters.getThreshold(),
```
originalname51: I like passing in a whitelist with the matcher params!

gtas-parent/gtas-commons/src/main/java/gov/gtas/services/matcher/MatchingServiceImpl.java
```diff
@@ -361,9 +411,8 @@ public int findMatchesBasedOnTimeThreshold(List<MessageStatus> messageStatuses)
         }
         endTime = System.nanoTime();
         int paxTotal = passengers == null ? 0 : passengers.size();
-        logger.debug("Passenger hit count and total run: " + totalMatchCount + " " + paxTotal);
-        logger.info("Execution time for performFuzzyMatching() for loop = " + (endTime - startTime) / 1000000
-                + "ms");
+        logger.debug("Passenger hit count and total run: {} {}",totalMatchCount, paxTotal);
+        logger.info("Execution time for performFuzzyMatching() for loop  = {} ms, {} passengers", (endTime - startTime) / 1000000, paxTotal);
```
originalname51: With the performance increases we are getting it might be worth it to switch the info here to debug.

gtas-parent/gtas-commons/src/main/java/gov/gtas/services/matcher/quickmatch/MatchingContext.java
```diff
@@ -298,46 +297,48 @@ private DerogResponse matchTraveler(HashMap<String, String> traveler) {
...
-								logger.debug(
-										"A Double Metaphone match and a Jaro Winkler distance hit of {} for traveler={}, derog={}.",
-										distance, traveler.get("full_name"), derogRecord.get("full_name"));
+						if (offset <= dobYearOffset) {
+							// calculate Jaro Winkler distance
+							Double distance = this.goodTextDistance(traveler.get("full_name"),
+									derogRecord.get("full_name"));
 
-								int offset = calculateYearOffset(traveler.get("DOB_Date"), derogRecord.get("DOB_Date"));
+							if (traveler.get("metaphones").equals(derogRecord.get("metaphones"))) {
```
originalname51: I think the if statements can be combined here. 
Logically 
traveler.get("metaphones").equals(derogRecord.get("metaphones"))
                                    && (distance >= jaroWinklerThreshold)

is equivalent here. (missing additional offset check as it is checked in first if statement).

gtas-parent/gtas-commons/src/main/java/gov/gtas/services/matcher/quickmatch/MatchingContext.java
```diff
@@ -298,46 +297,48 @@ private DerogResponse matchTraveler(HashMap<String, String> traveler) {
...
 
-								if (offset <= dobYearOffset) {
+								if (distance >= jaroWinklerThreshold) {
```
originalname51: Similar to above comment; I

I think 
                            } else if (distance >= jaroWinklerThreshold
                                    && (isSameDOBYearMonthDay(traveler.get("DOB_Date"), derogRecord.get("DOB_Date")))) {
is similar to the if statements below.

gtas-parent/gtas-commons/src/main/java/gov/gtas/services/matcher/quickmatch/MatchingContext.java
```diff
@@ -406,29 +407,24 @@ private DerogResponse matchTraveler(HashMap<String, String> traveler) {
...
-     * @throws ParseException
-     */
+	/**
+	 * 
+	 * @param travelerDate
+	 * @param deragDate
+	 * @return
+	 * @throws ParseException
+	 */
 	private int calculateYearOffset(String travelerDate, String deragDate) {
```
originalname51: Due to the performance needs of quick matcher I think two things could be considered here.

1. Move the formatter to the contructor as a class level variable. 
2. Use Java 8 libraries as they are significantly faster. The below code does the same thing but should be at least 2.5x faster.

	private int calculateYearOffset(String travelerDate, String deragDate) {
		DateTimeFormatter dateTimeFormatter = DateTimeFormatter.ofPattern("yyyy-MM-dd", Locale.ROOT);
        try {
            LocalDate travLocalDate = LocalDate.parse(travelerDate, dateTimeFormatter);
            LocalDate derogLocalDate = LocalDate.parse(deragDate, dateTimeFormatter);
            return Math.abs(travLocalDate.getYear() - derogLocalDate.getYear());
        } catch (Exception e) {
            logger.error("Failure to parse date");
        }
        return -1;
	}

originalname51: Also, I think the"isSameDobYearMonthDay" method could be reduced to 

	private boolean isSameDOBYearMonthDay(String travelerDate, String deragDate)  {
			return travelerDate.equalsIgnoreCase(deragDate);
	}

gtas-parent/gtas-commons/src/test/java/gov/gtas/services/QuickMatcherIT.java
```diff
@@ -0,0 +1,213 @@
+/*
```
originalname51: This class appears to be a copy of the "quickmatcherimpltest class and should probably be deleted.

tsegay: I agree. I thought "quickmatcherimpltest" was added to separate unit test from integration test. removed for now.

gtas-parent/gtas-commons/src/main/java/gov/gtas/repository/PaxWatchlistLinkRepository.java
tsegay: The "getPassengerMatchingInformation" makes sense. updated accordingly.

This is reusable outside the "MatchingServiceImpl". It has been updated to accept a list of passenger ids.

gtas-parent/gtas-commons/src/main/java/gov/gtas/services/matcher/MatcherParameters.java
tsegay: updated

gtas-parent/gtas-commons/src/main/java/gov/gtas/services/matcher/MatchingService.java
tsegay: imports organized!

gtas-parent/gtas-commons/src/main/java/gov/gtas/services/matcher/quickmatch/MatchingContext.java
tsegay: That would fail if the two dates are formatted differently.

gtas-parent/gtas-commons/src/main/java/gov/gtas/services/matcher/quickmatch/MatchingContext.java
tsegay: this was done to print different logs when jaro winkler matches but the dob fails but i guess just printing logs is not that useful.

gtas-parent/gtas-commons/src/main/java/gov/gtas/services/matcher/quickmatch/MatchingContext.java
originalname51: I do not think there is an expectation to get or make watchlist items with more than year-month-date, We control how the traveler bithdate is formatted within the class. Hopefully that is a case that while currently can be fed into the matcher it would not occur. Perhaps we can throw an exception if the date is not formatted as we expect and put a requirement on the class to get dates formatted in yyyy-MM-dd.

Depending on the performance gains from using dateTimeFormatter this might be an acceptable requirement to put on the class - it would also open up the string comparison for exact date match which is likely faster than instantiating objects to immediately compare and discard them.

PR #1201, opened 2019-06-14
Flights grid

gtasautumn:
data sync

### DIFF
--- a/gtas-parent/gtas-commons/src/main/java/gov/gtas/repository/FlightRepositoryImpl.java
+++ b/gtas-parent/gtas-commons/src/main/java/gov/gtas/repository/FlightRepositoryImpl.java
@@ -15,9 +15,11 @@ import gov.gtas.services.dto.FlightsRequestDto;
 import gov.gtas.services.dto.SortOptionsDto;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Date;
 import java.util.List;
 import java.util.Optional;
+import java.util.*;
 
 import javax.persistence.EntityManager;
 import javax.persistence.PersistenceContext;
@@ -69,6 +71,7 @@ public class FlightRepositoryImpl implements FlightRepositoryCustom {
 		Join<Flight, FlightPassengerCount> passengerCountJoin = root.join("flightPassengerCount", JoinType.LEFT);
 		Join<Flight, MutableFlightDetails> mutableFlightDetailsJoin = root.join("mutableFlightDetails", JoinType.LEFT);
 		Predicate etaCondition = getETAPredicate(dto, cb, mutableFlightDetailsJoin);
+		Predicate updCondition = getUpdatedAfterPredicate(dto, cb, root);
 
 		// sorting
 		if (dto.getSort() != null) {
@@ -166,6 +169,9 @@ public class FlightRepositoryImpl implements FlightRepositoryCustom {
 		if (etaCondition != null) {
 			predicates.add(etaCondition);
 		}
+		if (updCondition != null) {
+			predicates.add(updCondition);
+		}
 		q.select(root).where(predicates.toArray(new Predicate[] {}));
 		TypedQuery<Flight> typedQuery = em.createQuery(q);
 
@@ -177,7 +183,7 @@ public class FlightRepositoryImpl implements FlightRepositoryCustom {
 		typedQuery.setMaxResults(dto.getPageSize());
 
 		// Runs exact query above without limits to get a count.
-		long count = getCountOfQuery(dto, cb, countPredicates);
+		long count = getCountOfQuery(dto, cb, countPredicates, root);
 
 		logger.debug(typedQuery.unwrap(org.hibernate.Query.class)
 				.getQueryString());
@@ -200,13 +206,18 @@ public class FlightRepositoryImpl implements FlightRepositoryCustom {
 		}
 	}
 
-	private long getCountOfQuery(FlightsRequestDto dto, CriteriaBuilder cb, List<Predicate> countPredicates) {
+	private long getCountOfQuery(FlightsRequestDto dto, CriteriaBuilder cb, List<Predicate> countPredicates, Root<Flight> root) {
 		CriteriaQuery<Long> countQuery = cb.createQuery(Long.class);
 		Root<Flight> countRoot = countQuery.from(Flight.class);
 		Join<Flight, MutableFlightDetails> countMutableFlightsInfoJoin = countRoot.join("mutableFlightDetails", JoinType.LEFT);
 		Predicate countEtaCondition = getETAPredicate(dto, cb, countMutableFlightsInfoJoin);
-		if (countEtaCondition != null) {
+		Predicate countUpdCondition = getUpdatedAfterPredicate(dto, cb, root);
+
+    if (countEtaCondition != null) {
 			countPredicates.add(countEtaCondition);
+		}
+		if (countUpdCondition != null) {
+			countPredicates.add(countUpdCondition);
 		}
 		countQuery.select(cb.count(countRoot)).where(
 				countPredicates.toArray(new Predicate[]{}));
@@ -228,6 +239,23 @@ public class FlightRepositoryImpl implements FlightRepositoryCustom {
 			etaCondition = cb.and(startPredicate, endPredicate);
 		}
 		return etaCondition;
+	}
+
+	private Predicate getUpdatedAfterPredicate(FlightsRequestDto dto, CriteriaBuilder cb, Root<Flight> root) {
+//    Set<String> foo = new HashSet<>(Arrays.asList("ET"));
+//
+//    Predicate uaCondition = root.get("carrier").in(foo);
+     Predicate uaCondition = null;
+     Date minDate = dto.getUpdatedAfter();
+    
+		  if (minDate != null) {
+       uaCondition = cb.or(
+         cb.greaterThan(root.get("createdAt"), minDate),
+         cb.and(
+           cb.isNotNull(root.get("updatedAt")),
+           cb.greaterThan(root.get("updatedAt"), minDate) ));
+		  }
+		return uaCondition;
 	}
 
 	@Override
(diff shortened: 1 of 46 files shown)


gtas-parent/gtas-commons/src/main/java/gov/gtas/repository/FlightRepositoryImpl.java
```diff
@@ -230,6 +241,23 @@ private Predicate getETAPredicate(FlightsRequestDto dto, CriteriaBuilder cb, Joi
 		return etaCondition;
 	}
 
+	private Predicate getUpdatedAfterPredicate(FlightsRequestDto dto, CriteriaBuilder cb, Root<Flight> root) {
+//    Set<String> foo = new HashSet<>(Arrays.asList("ET"));
```
originalname51: Nitpick but I think this could be beneficial with commented out test code and the method code formatted.

gtas-parent/gtas-commons/src/main/java/gov/gtas/services/FlightServiceImpl.java
```diff
@@ -25,6 +25,7 @@
 import javax.persistence.PersistenceContext;
 
 import org.apache.commons.lang3.tuple.Pair;
+import org.apache.lucene.queries.payloads.MaxPayloadFunction;
```
originalname51: Import statement does not appear to be used.

gtas-parent/gtas-commons/src/main/java/gov/gtas/services/FlightServiceImpl.java
```diff
@@ -66,8 +67,24 @@ public FlightsPageDto findAll(FlightsRequestDto dto) {
...
 		Pair<Long, List<Flight>> tuple2 = flightRespository.findByCriteria(dto);
 		List<Flight> flights = tuple2.getRight();
-		List<FlightVo> vos = convertFlightToFlightVo(flights);
-		return new FlightsPageDto(vos, tuple.getLeft());
+    List<FlightVo> vos = convertFlightToFlightVo(flights);
+    Date lastUpdated = getLastUpdated(flights);
+		return new FlightsPageDto(vos, tuple.getLeft(), lastUpdated, false);
+    }
+
+    private Date getLastUpdated(List<Flight> flights) {
```
originalname51: Method has some commented out code that would be beneficial to be removed. 

From a style point of view I think it would be better to keep '{' and '}' on if statements as this is more commonly used on the app code.

gtas-parent/gtas-webapp/src/main/webapp/flights/FlightsController.js
```diff
@@ -107,36 +108,61 @@
...
 
+
+      var clearRefresh = function() {
+        console.log('refresh cleared');
+        $scope.model.updatedAfter = null;
+        clearInterval(refresher);
+      }
+
+      var startRefresh = function() {
+        refresher = setInterval(fetchMethods.flightsRefresh, 30000);
```
originalname51: Is it worth having the flightsRefresh method turn the interval off until it has completed? Currently if the server does not respond within 30 seconds another request will be sent.

gtas-parent/gtas-commons/src/main/java/gov/gtas/services/FlightServiceImpl.java
```diff
@@ -66,8 +67,24 @@ public FlightsPageDto findAll(FlightsRequestDto dto) {
...
+      if (flights.isEmpty()) return null;
+
+      // Date maxcreated = flights.stream().max(Comparator.comparing(v -> v.getCreatedAt())).getCreatedAt();
+      // Date maxupdated = flights.stream().max(Comparator.comparing(v -> v.getUpdatedAt())).getUpdatedAt();
+
+      Date latest = flights.get(0).getCreatedAt();
```
originalname51: If we assume the first query is the most recent query then a timestamp of now would be a good value for updatedAfter. When passing to the back-end we would be sure to only be asking for flights/dates that are 100% new with no calculation required.

gtas-parent/gtas-webapp/src/main/webapp/flights/FlightsController.js
```diff
@@ -107,36 +108,61 @@
                   $scope.queryLimitReached = flights.data.result.queryLimitReached;
               }
               else{
-                  setFlightsGrid($scope.flightsGrid, flights || {flights: [], totalFlights: 0});
+                setFlightsGrid($scope.flightsGrid, flights || {flights: [], totalFlights: 0});
+                $scope.model.updatedAfter = flights.data.lastUpdated || $scope.model.updatedAfter;
```
originalname51: This mirrors a comment made on FlightServiceImpl. It would guarantee only updated or newly created records be included in a update statement by searching for all flights after the point in time which the query was executed instead of the point in time the latest flight was updated.

gtas-parent/gtas-webapp/src/main/webapp/flights/FlightsController.js
```diff
@@ -107,36 +108,61 @@
...
+            }
           },
           resolvePage = function () {
               populateAirports();
               fetchMethods[stateName]();
           };
 
+
+      var clearRefresh = function() {
+        console.log('refresh cleared');
```
originalname51: This might warrant removal as a debug statement.

gtas-parent/gtas-webapp/src/main/webapp/flights/FlightsController.js
```diff
@@ -192,6 +218,31 @@
           $scope.model.dest = destinationAirports;
       };
 
+      //APB - refac
+      function merge(data) {
+        console.log(`refreshed: ${data.data.flights.length} records`);
```
originalname51: This might warrant removal as a debug statement.

gtas-parent/gtas-webapp/src/main/webapp/flights/FlightsController.js
```diff
@@ -192,6 +218,31 @@
...
+            merged[flight.id] = flight;
+          });
+        }
+
+        data.data.flights.map(function (newflight) {
+          merged[newflight.id] = newflight;
+        });
+
+        flights.data.flights = Object.values(merged);
+        flights.data.totalFlights = flights.data.flights.length;
```
originalname51: The totalFlights is the result of the count query which is typically larger than the length of the array of flights returned.

gtas-parent/gtas-webapp/src/main/webapp/flights/FlightsController.js
```diff
@@ -192,6 +218,31 @@
...
+          flights.data.flights.map(function (flight) {
+            merged[flight.id] = flight;
+          });
+        }
+
+        data.data.flights.map(function (newflight) {
+          merged[newflight.id] = newflight;
+        });
+
+        flights.data.flights = Object.values(merged);
```
originalname51: Given a condition of an updated flight coming back no other flight gets "bumped" from the list which could cause an "overrun" of items per page. 

Is there a guarantee of sorting here?

gtas-parent/gtas-commons/src/main/java/gov/gtas/repository/FlightRepositoryImpl.java
originalname51: The predicate generates the following sql:

            and (
                flight0_.created_at>? 
                or (
                    flight0_.updated_at is not null
                ) 
                and flight0_.updated_at>?
            )

originalname51: The flight is never supposed to be updated, instead the flight mutable details is what is supposed to be updated due to concurrency.

I recommend an update check on mutable flight details instead of the flight, although this will not check for cases/hits/or passenger count.

gtasautumn: Adding a ticket to include update data on mutable tables.

gtas-parent/gtas-commons/src/main/java/gov/gtas/services/FlightServiceImpl.java
gtasautumn: Noted, thanks!

gtas-parent/gtas-commons/src/main/java/gov/gtas/services/FlightServiceImpl.java
gtasautumn: Agreed, on some sorts the most recent updated recs will not be relevant to the user's current view. Switching to a simpler refetch will resolve this though I had hoped to avoid the additional data.

gtas-parent/gtas-webapp/src/main/webapp/flights/FlightsController.js
gtasautumn: I like that idea. Might make sense to only queue up the next refresh after a successful one until we move away from polling. I'll also look at an autokill+notification on server error. Thanks.

gtas-parent/gtas-commons/src/main/java/gov/gtas/repository/FlightRepositoryImpl.java
```diff
@@ -200,12 +202,13 @@ static void generateFilters(FlightsRequestDto dto, CriteriaBuilder cb, List<Pred
 		}
 	}
 
-	private long getCountOfQuery(FlightsRequestDto dto, CriteriaBuilder cb, List<Predicate> countPredicates) {
+	private long getCountOfQuery(FlightsRequestDto dto, CriteriaBuilder cb, List<Predicate> countPredicates, Root<Flight> root) {
```
originalname51: The parameter root was added but is not used - likely a hold-over from previous modification.

gtas-parent/gtas-commons/src/main/java/gov/gtas/services/FlightServiceImpl.java
```diff
@@ -66,8 +65,8 @@ public FlightsPageDto findAll(FlightsRequestDto dto) {
 		flightRespository.flush();
 		Pair<Long, List<Flight>> tuple2 = flightRespository.findByCriteria(dto);
 		List<Flight> flights = tuple2.getRight();
-		List<FlightVo> vos = convertFlightToFlightVo(flights);
-		return new FlightsPageDto(vos, tuple.getLeft());
+    List<FlightVo> vos = convertFlightToFlightVo(flights);
+		return new FlightsPageDto(vos, tuple.getLeft(), false);
```
originalname51: The constructor here is changed but the class it references is not. This is probably another hold-over from updates.

gtas-parent/gtas-commons/src/main/java/gov/gtas/services/dto/FlightsRequestDto.java
```diff
@@ -40,6 +40,7 @@
 	private Date etaStart;
 	@JsonFormat(shape = JsonFormat.Shape.STRING, pattern = DATE_FORMAT)
 	private Date etaEnd;
+	@JsonFormat(shape = JsonFormat.Shape.STRING, pattern = DATE_FORMAT)
```
originalname51: This is a holdover from updatedAfter getting deleted.

Issue #1056, opened 2019-04-04
Configure JNDI Connection Pool


tsegay:
GTAS has been updated (667538cf2098a51ed32c6263a6b02162d33754af) to use JNDI connection pool to connect to the database. This will make it easy to swap connection string information without the need to build/deploy the application. 

Complete the steps below to configure JNDI  

1. Configure Datasource in tomcat – (`gtas-parent/gtas-configs/servers/tomcat/server.xml` is preconfigured with the default connection string). Copy the file into `$TOMCAT_HOME/conf/server.xml`

2. Configure ResourceLink – (`gtas-parent/gtas-configs/servers/tomcat/context.xml` is preconfigured). Copy the file into `$TOMCAT_HOME/conf/context.xml`

~~3. You might need to copy mariadb java client jar into tomcat's lib folder (`$TOMCAT_HOME/lib`). After building the application, the jar can be found in `gtas-webapp/target/gtas/WEB-INF/lib/mariadb-java-client-2.3.0.jar`~~

tsegay:
step #-3 is not needed since the same jar is already included inside the applications lib folder.

PR #1203, opened 2019-06-18
Reintroduce hibernate properties

originalname51:
I'd like to reintroduce the database management to hibernate.properties as we will be able to run more integration test and see 

I see some pretty good pros here including the following:
** Allowing for integration test to run which were broken in #1056. Example includes the PNR integration test.

** Allowing application configuration of database to be managed by hibernate and clearly configured within the application code

** One place to look for all application configuration instead of multiple


originalname51:
Did not mean to request merge into master.

PR #1204, opened 2019-06-18
Reintroduce c3p0/hibernate properties

originalname51:
I'd like to reintroduce the database management to hibernate.properties as we will be able to run more integration test and see

I see some pretty good pros here including the following:
** Allowing for integration test to run which were broken in #1056. Example includes the PNR integration test.

** Allowing application configuration of database to be managed by hibernate and clearly configured within the application code

** One place to look for all application configuration instead of multiple


TeddyTesfaye:
Now that we have multiple environments, I still prefer to keep the JNDI configuration for ease of deployment because the connection info always has to stay in the server and we do not have to set the database connection information for each build or do not have to change the database connection info when we deploy in dev environment, test environment, etc..

originalname51:
Confirming that unit test do work. Functionally the result is the same, if advantages are found pulling it in the code we can do so at that time. Closing PR.

originalname51:
After further review c3p0 libraries have easy configuration, especially of wait_timeout , which when set to lower values can create issues.

originalname51:
Another difference between c3p0 and datasource is when the database connection is killed (say the database is restarted) the applications do not recover. c3p0 will automatically recover the connection pool.

gtasautumn:
Is this related to the issue where running the app locally, you sometimes
lose db connectivity? Happens more now that I have mariadb in docker, but
havent pinned down why it happens yet, though in .net its to do with how
the pool is set to recycle.

tsegay:
@gtasautumn, you are right. This is an issue with how tomcat manages connection pools, unable to recycle idle connections properly.  I just tested the configuration below locally and It works fine (with local mariadb or on docker) if `maxIdle` is set to 0 or 1 which means every connection returned to the pool would be closed. That kind of defeats the idea of using connection pool but still keeps the datasource decoupled from the application source code. 

```
    <Resource 
    	auth="Container" 
    	driverClassName="org.mariadb.jdbc.Driver" 
    	logAbandoned="true" 
    	maxIdle="0" 
    	maxTotal="200" 
    	maxWaitMillis="10000" 
    	name="jdbc/gtasDataSource" 
    	password="admin" 
    	removeAbandonedOnBorrow="true" 
    	removeAbandonedOnMaintenance="true" 
    	removeAbandonedTimeout="60" 
    	testOnBorrow="true" 
    	type="javax.sql.DataSource" 
    	url="jdbc:mariadb://localhost:3306/gtas" 
    	username="root" 
    	validationQuery="select 1" 
    />
```

originalname51:
@tsegay the above configuration does not reconnect to the database if the database connection is turned off and back on.

To reproduce turn on GTAS, log on, log out, turn off database, turn on database, and you will no longer be able to log in.

tsegay:
I could not reproduce it. I'm running mariadb on docker. I followed the steps below and it worked fine.

1. run GTAS -> log in , log out 
2. remove  mariadb container
3. run the container again 
4. was able to log in

originalname51:
After review there appears to be configuration differences between Windows/Linux/Docker. As these issues do not exist on c3p0 there will be a switch back.

Configuring database pools through JNDI will be nice to have for the deployment of multiple GTAS loaders / rule runners when they are eventually split up. 

https://github.com/US-CBP/GTAS/issues/1216  has been made to capture this requirement.

PR #1207, opened 2019-06-19
visualize itinerary of a single flight 

tsegay:
see commit message for the updates. I couldn't easy hide the "add a filter" link and the text below. 
Screen Shot 2019-06-19 at 1 03 51 PM


originalname51:
pr will be re-opened in future; closing .

PR #1231, opened 2019-07-08
Scripts for installing Neo4j/pentaho

kasimhelil:


gtas-parent/scripts/graph_db/mac/neo4j/install.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+PARENT_DIR=$( cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P )
+
+#A directory where neo4j be installed
+cd /usr/local/bin/
+
+#Download neo4j
+wget http://dist.neo4j.org/neo4j-community-3.5.3-unix.tar.gz
```
tsegay: This will fail if the current user does not have write permission on `/opt`.

you might need to set the correct permission before line 8, something like ...

`sudo chmod u+rw /opt`

kasimhelil: Thank you @tsegay for reviewing the code. On mac `<neo4j>` will be installed under `</usr/local/bin/>`, not `</opt>`. And I will add `<sudo chmod u+rw /opt>` for linux installation.

PR #1279, opened 2019-08-01
Dockerizing neo4j

kasimhelil:


gtas-neo4j-etl/.env
```diff
@@ -0,0 +1,2 @@
+LOCAL_DRIVE_MAPPING_INPUT=../__data/gtas_in
```
originalname51: I recommend a portion of the readme dedicated to setting up the in and out folder.

tsegay: The instruction is included in the [Docker Usage Instructions.md](https://github.com/US-CBP/GTAS/blob/master/Docker%20Usage%20Instructions.md#configure-local-folders)

Issue #1258, opened 2019-07-30
Parser Tuning | more phone numbers


dje04001:
similar to issue #1257 , parent level

ADD++700:::::::EK CTC H 1234567890'
ADD++700:::::::EK CTC B 0987654321'
ADD++700:::::::EK CTC M 5432109876'

H = Home
B = Business
M = Mobile

At individual passenger level, in the SSR segment

SSR+CTCM:HK::YY::::::0123456789+::001'

originalname51:
#1295 is pr

lnpeng:
I uploaded to the dev boxes the messages that have the following address segments:
```
ADD++700:::::::EK CTC H 1234567890'
ADD++700:::::::EK CTC B 0987654321'
ADD++700:::::::EK CTC M 5432109876'
```
The messages were parsed correctly and displayed in the Phone Numbers table of the Passenger Details view.

PR #1295, opened 2019-08-06
#1258 Parser Tuning on ADD phone numbers.

originalname51:
ADD was not to specification. The specification says we should parse phone numbers from free text.

It is suspected ADD was written to parse PNR 11.1 but we do not have a spec doc.

Updated the ADD to parse for (**suspected**)11.1 and 13.1 PNR.

Added unit testing. Made it clear that the numbers we are parsing are both fake AND obviously phone numbers.


gtas-parent/gtas-parsers/src/main/java/gov/gtas/parsers/pnrgov/segment/ADD.java
```diff
@@ -50,71 +50,95 @@
...
-                    if(StringUtils.isNotBlank(temp)){
-                    	String[] tokens=temp.split(" ");
-                    	if(tokens.length >= 3){
-                    		this.city = tokens[0];
-                    		this.stateOrProvinceCode = tokens[1];
-                    		this.postalCode = tokens[2];
-                    	}
+        }
+        if (StringUtils.isBlank(c.getElement(0)) && StringUtils.isNotBlank(c.getElement(7))
+                && c.getElement(7).contains("TBM")) {
```
tsegay: just to be consistent ... use the constant variables in places of the magic numbers/index.

gtas-parent/gtas-parsers/src/main/java/gov/gtas/parsers/pnrgov/segment/ADD.java
```diff
@@ -50,71 +50,95 @@
...
+            //ADD++:::::::TBM MAIL TO+:::::::FIRST NAME LAST NAME:::::::99 STREET:::::::CITY STATE LONGZIPCODE'
+            Composite c2 = getComposite(2);
```
tsegay: same comment as above.

gtas-parent/gtas-parsers/src/main/java/gov/gtas/parsers/pnrgov/segment/ADD.java
```diff
@@ -50,71 +50,95 @@
...
+            if (c2 != null) {
+                this.addressType = "MAIL TO";
+                this.streetNumberAndName = c2.getElement(14);
```
tsegay: same comment as above.

PR #1297, opened 2019-08-06
1262

### LINKED ISSUE #1262: Hand Null values when pointing to front end
PNR related, addresses (city) are null and result in a null pointer.

kasimhelil:
Resolves #1262

### DIFF
--- a/gtas-parent/gtas-webapp/src/test/java/gov/gtas/controller/PassengerDetailsControllerUT.java
+++ b/gtas-parent/gtas-webapp/src/test/java/gov/gtas/controller/PassengerDetailsControllerUT.java
@@ -1,0 +1,31 @@
+package gov.gtas.controller;
+
+import gov.gtas.vo.passenger.AddressVo;
+import gov.gtas.vo.passenger.PnrVo;
+import org.junit.Test;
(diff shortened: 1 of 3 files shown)


gtas-parent/gtas-webapp/src/test/java/gov/gtas/controller/PassengerDetailsControllerUT.java
```diff
@@ -0,0 +1,31 @@
+package gov.gtas.controller;
+
+import gov.gtas.vo.passenger.AddressVo;
+import gov.gtas.vo.passenger.PnrVo;
+import org.junit.Test;
+import static org.junit.Assert.*;
+import java.util.ArrayList;
+
+public class PassengerDetailsControllerUT {
```
originalname51: In order to run this with other unit test and to keep with how other files are named please append "Test" to the end of the name instead of "UT".
Please change the name of this PassengerDetailsControllerUT to PassengerDetailsControllerTest

See https://maven.apache.org/surefire/maven-surefire-plugin/examples/inclusion-exclusion.html

originalname51:
Looks good.

Issue #1196, opened 2019-06-13
Neo4J Tab on Passenger Page


originalname51:
Add a tab to the passenger page with the results of a cipher query.

- [ ] #1197 Front End
- [x] #1198 Back End

originalname51:
Solution will be fully on front end.

PR #1316, opened 2019-08-19
Neo4j visual

gtasautumn:
Issues #1197 and #1196, display neo4j data in the paxdetail page for canned queries against the current flight and pax info.


gtas-parent/gtas-job-scheduler-war/src/main/resources/singleNodeConfig.json
```diff
@@ -11,7 +11,7 @@
     "password": null,
     "subscriptionsPerConnection": 5,
     "clientName": null,
-    "address": "redis://0.0.0.0:6379",
+    "address": "redis://127.0.0.1:6379",
```
originalname51: Should this be deleted? Redis was removed in v1.3.

gtas-parent/gtas-commons/src/main/resources/commonservices.properties
```diff
@@ -31,6 +31,8 @@ activemq.broker.url = tcp://localhost:61616
 #########
 ### REDIS ###
 redis.connection.string=redis://0.0.0.0:6379
+### REDIS (windows) ###
```
originalname51: Should this be deleted? Redis was removed in 1.3

gtas-parent/gtas-loader/src/main/java/gov/gtas/services/GtasLoaderImpl.java
```diff
@@ -270,10 +269,10 @@ public Flight processFlightsAndBookingDetails(List<FlightVo> flights, Set<Flight
                 BeanUtils.copyProperties(fvo, mfd, getNullPropertyNames(fvo));
                 mfd.setEtaDate(DateUtils.stripTime(mfd.getEta()));
                 if (mfd.getEtd() == null) {
-                    // Special case where prime flight doesnt have a timestamp
+                  Date primeFlightTimeStamp = new Date(Long.parseLong(primeFlightKey[ETD_DATE_WITH_TIMESTAMP]));
+                  // Special case where prime flight doesnt have a timestamp
                     //  we use the prime TDT with the 125 or 87 LOC code \ 189 DTM code for APIS
                     //  or tvl level 0 etd timestamp for PNR if this is the case.
```
originalname51: having the comment describe the code under it is slightly more readable to me.

gtas-parent/gtas-job-scheduler-war/src/main/resources/singleNodeConfig.json
gtasautumn: shd go away in the merge

gtas-parent/gtas-webapp/src/main/webapp/pax/PaxService.js
```diff
@@ -4,262 +4,299 @@
...
-                          file: file
-                        }
-                    }).progress(function (evt) {
-                    	//progress tracker potentially
-                    }).success(function (data, status, headers, config) {
-                        return "success";
-                    });
-            	}
-            };
+          function getAddressPath() {
```
originalname51: I think the following code should be commented to explain to maintainers what the numbers mean.

gtas-parent/gtas-webapp/src/main/webapp/pax/pax.detail.html
```diff
@@ -4,856 +4,910 @@
...
-                  </span>
+                </span>
               </td>
-              <td>{{ passenger.flightETA | date:'yyyy-MM-dd HH:mm' }}</td>
             </tr>
           </table>
-        </h6>
-      </div>
+         
+         <div class="map-wrapper" ng-if="kibanaSettings.dashboardDisabled == false">
```
originalname51: dashboardDisabled  has been replaced with default.landing.page. This will need to be re-added if needed.

gtas-parent/gtas-webapp/src/main/webapp/pax/pax.detail.html
```diff
@@ -868,106 +922,176 @@ <h4 class="h-label no-margin-top" >Current Itinerary</h4>
...
+                                        <td><span class="ppt-label" title="All passenger hits for this flight">All Hits</span></td>
+                                      </tr>
+                                      <tr id="Document" ng-click="onClickSavedGraph('documentall')">
+                                          <td><svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 448 512" role="img" class="pptdocument">
+                                            <path d="{{documentPath()}}"/></svg></td>
+                                      <td><span class="ppt-label" title="All passenger documents for this flight">All Documents</span></td>
+                                          </tr>
+                                        </table>
+                                </div>
+                                <!-- <div id="popoto-taxonomy">
```
originalname51: Is commented out code needed?

gtas-parent/gtas-webapp/src/main/webapp/pax/pax.detail.html
```diff
@@ -868,106 +922,176 @@ <h4 class="h-label no-margin-top" >Current Itinerary</h4>
...
+                                <div id="popoto-graph" class="ppt-div-graph">
+                                    <!-- Graph will be generated here-->
+                                </div>
+                            </div>
+                        
+                            <div id="popoto-query" class="ppt-container-query">
+                                <!-- Query viewer will be generated here -->
+                            </div>
+                        
+                            <!-- <div id="popoto-cypher" class="ppt-container-cypher">
```
originalname51: Is commented out coded required here?

gtas-parent/gtas-webapp/src/main/java/gov/gtas/controller/ConfigurationController.java
```diff
@@ -53,4 +65,18 @@ public String getNeo4J() {
         return neo4jUrl;
     }
 
+    @RequestMapping(method = RequestMethod.GET, value = "/api/config/cypherUrl",  produces="text/plain")
+    public String getCypherUrl() {
+        return cypherUrl;
+    }
+
+    @RequestMapping(method = RequestMethod.GET, value = "/api/config/cypherAuth",  produces="text/plain")
```
originalname51: it looks like we are encoding the password instead of encrypting it. 

How is security being handled? e.g. How do we handle the case of an individual hitting the endpoint and decoding the resulting text string to find the username / password to neo4j?

originalname51:
I notice there are several conflicts on this branch. I suggest merging from dev to neo4j-visual to resolve these.

gtas-parent/gtas-job-scheduler-war/src/main/resources/singleNodeConfig.json
originalname51: It will be changed in the merge from 0.0.0.0:6379 to 127.0.0.1:6379. I think changes here are likely outside of the scope of the tickets, but if we are going to make a change to remove it instead of change the string.

originalname51:
The java code has no unit test. Please add at least 1 unit test (in an area such as hashing).

gtas-parent/gtas-webapp/src/main/java/gov/gtas/controller/PassengerDetailsController.java
```diff
@@ -123,6 +124,7 @@ public PassengerVo getPassengerByPaxIdAndFlightId(
 			bagList = new ArrayList<>(bagRepository.findFromFlightAndPassenger(flight.getId(), t.getId()));
 		}
 		vo.setPaxId(String.valueOf(t.getId()));
+		vo.setPaxIdTag(t.getPassengerIDTag().getIdTag());
```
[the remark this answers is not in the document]
gtasautumn: null check is added on the pax idtag. Not sure I agree with the logic there even in the case of missing data, and where it is null there will be no way to track that pax in the graphdb, so the systems will be out of sync.

gtas-parent/gtas-loader/src/main/java/gov/gtas/services/LoaderUtils.java
```diff
@@ -173,11 +173,13 @@ public Flight createNewFlight(FlightVo vo, String [] primeFlightKey) throws Pars
         Override any data that does not match the prime flight key.
         This can happen in the case of a code share flight.
          * */
-        f.setDestination(primeFlightKey[GtasLoaderImpl.PRIME_FLIGHT_DESTINATION]);
-        f.setOrigin(primeFlightKey[GtasLoaderImpl.PRIME_FLIGHT_ORIGIN]);
-        f.setEtdDate(new Date(Long.parseLong(primeFlightKey[GtasLoaderImpl.ETD_DATE_NO_TIMESTAMP_AS_LONG])));
-        f.setCarrier(primeFlightKey[GtasLoaderImpl.PRIME_FLIGHT_CARRIER]);
-        f.setFlightNumber(primeFlightKey[GtasLoaderImpl.PRIME_FLIGHT_NUMBER_STRING]);
+
```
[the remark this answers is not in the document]
gtasautumn: The current change is correct. It's more readable, it passes in only the fields it uses, and it doesn't need access to another class for information on where in the array its data is.

In general, its bad design any time a util requires this kind of reference. A util is for generic shared functionality and should not need to share implementation details like this, especially for something so simple. If it really requires another service or entity, or if it won't be used outside that class, it should be encapsulated by that class. Otherwise, it shouldn't care whose data its working on.

On the formatting note, agreed. My ide autoformats on save, it's causing me issues as well. We should discuss as a team some general defaults, see what works.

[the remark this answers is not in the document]
originalname51: I suggest the prime flight key string array to be wrapped by an object that in turn can return these values.

If we are refactoring this code then I think it really belongs as a static method that constructs a flight in the flights class that takes the prime flight wrapper object and returns a flight.

This solves the coupling issue between gtasloaderimpl constant values (which was a substitute for a class wrapper), allows the flight class encapsulation, and stops us from introducing a method with 6 parameters.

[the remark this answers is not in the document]
gtasautumn: It should absolutely be static, as should most of the methods in that util. But it looks like only GtasLoaderImpl even has a reference to this function, so they were both referencing each other for a function that belonged in the loader. I don't know the other loader code, whether they have similar "createflight" functionality. If they do we can consolidate where possible and make a cleaner generic function. If they don't create flights, we can just move it into the loader.

gtas-parent/gtas-parsers/src/test/resources/redisloader.properties
```diff
@@ -9,5 +9,5 @@ inbound.loader.jms.queue=GTAS_INBOUND_Q_REDIS
 outbound.loader.jms.queue=GTAS_LOADER_Q
 #########
 ### REDIS ###
-redis.connection.string=redis://0.0.0.0:6379
+redis.connection.string=redis://127.0.0.1:6379
```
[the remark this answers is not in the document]
gtasautumn: So the whole file is for the redisloader, is it still needed? We can put in a ticket to drop all the redis files in test. SingleNodeConfig.json and .yaml also still have redis references.

gtas-parent/gtas-loader/src/main/java/gov/gtas/services/LoaderUtils.java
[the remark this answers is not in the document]
originalname51: I do not think object creation belongs in a util class. I think the cleanest way would be a static method along the lines of the following signatures: 

    public static Flight fromPrimeFlightKey(String [] primeFlightKey) {

or

    public static Flight fromPrimeFlightKey(PrimeFlightKey primeFlightKey) {

That way a flight can be generated by saying

Flight flight = Flight.fromPrimeFlightKey(primeFlightKey)

gtas-parent/gtas-webapp/src/main/webapp/main.html
```diff
@@ -230,6 +228,12 @@
   <script src="resources/bower_components/trix/dist/trix.js"></script>
   <script src="resources/bower_components/angular-trix/dist/angular-trix.min.js"></script>
   <script src='seatsMap/jquery.seat-charts.min.js'></script>
+
+  <script src='resources/node_modules/d3/dist/d3.min.js'></script>
+  <script src='resources/node_modules/vaquita/dist/vaquita.js'></script>
+  <link rel="stylesheet" href='resources/node_modules/vaquita/dist/vaquita.min.css' type="text/css" media="all" />
```
tsegay: style sheets should be linked in the header.

gtas-parent/scripts/db/rollback/rollback.sql
```diff
@@ -0,0 +1,4 @@
+-- Restore FLIGHT table
+ALTER TABLE `FLIGHT`
+	DROP COLUMN IF EXISTS `id_tag`;
+
```
tsegay: does dropping the column also drops the index?

tsegay: It does. [Mariadb Doc](https://mariadb.com/kb/en/library/alter-table/#drop-column) 

> The index will be dropped if all columns from the index were dropped

originalname51:
The default.application.properties and application.properties.EXAMPLE  kibana link is needed for the dashboard. The hard coded kibana link removed in the 75dad5c92eb7cb8c6702f6abf7881918c3361cb0 commit looks good. 

Please revert the changes to default.application.properties and application.properties.EXAMPLE.

Otherwise looks good!

TeddyTesfaye:
During testing, I found out the graph related to a passenger do not get loaded the first time but it loads if the page or the section that shows the graph is refreshed. A separate ticket should be created to address this issue, i.e, the issue related to displaying graphs on initial page load. Based on my initial investigation of the issue, it looks like it is related to loading the correct version of js libraries (nothing related to the actual code). It requires a further investigation.

PR #1317, opened 2019-08-20
Added fixes to user location

TeddyTesfaye:
@kasimhelil

### DIFF
--- a/gtas-parent/gtas-webapp/src/main/java/gov/gtas/common/UserLocationSetting.java
+++ b/gtas-parent/gtas-webapp/src/main/java/gov/gtas/common/UserLocationSetting.java
@@ -26,8 +26,77 @@ public class UserLocationSetting {
 	@Autowired
 	private AppConfigurationService appConfigurationService;
 
-	public void setPrimaryLocation(HttpServletRequest httpServletRequest, String userId) {
+	public UserLocationStatus setPrimaryLocation(HttpServletRequest httpServletRequest, String userId) {
+
+		UserLocationStatus userLocationStatus = getPrimaryLocationStatus(httpServletRequest, userId);
+
+		/*
+		 * The location exists and is also a primary one is enabled in GTAS
+		 * user_location table so just set the http session
+		 */
+		if (userLocationStatus.getPrimaryLocationAirport() != null && userLocationStatus.isPrimaryLocationCreated()
+				&& userLocationStatus.isPrimaryLocationEnabledInDb()) {
+			httpServletRequest.getSession().setAttribute(Constants.USER_PRIMARY_LOCATION,
+					userLocationStatus.getPrimaryLocationAirport());
+			logger.info(
+					"So user primary location exists so the user Location is just set in the HTTP session successfull");
+		}
+
+		else if (userLocationStatus.getPrimaryLocationAirport() != null && userLocationStatus.isPrimaryLocationCreated()
+				&& !userLocationStatus.isPrimaryLocationEnabledInDb()) {
+			/*
+			 * One or more location airports exist but none of them are marked primary in
+			 * GTAS user_location table so mark one of them as primary location and set the
+			 * http session
+			 */
+			try {
+				userLocationService.updateUserPrimaryLocation(userId, userLocationStatus.getPrimaryLocationAirport(),
+						true);
+				logger.info("One of the user's location airports is set as a primary location in user_location table");
+			} catch (Exception e) {
+				logger.error("An error has occurred when updating user location in the database.");
+			}
+
+		} else if (userLocationStatus.getPrimaryLocationAirport() != null
+				&& !userLocationStatus.isPrimaryLocationCreated()
+				&& !userLocationStatus.isPrimaryLocationEnabledInDb()) {
+			/*
+			 * No location airports were found in user_location table so insert the dashboard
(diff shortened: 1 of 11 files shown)


originalname51:
I think a unit test under setPrimaryLocation would be beneficial.

gtas-parent/gtas-webapp/src/main/java/gov/gtas/controller/FlightPassengerController.java
```diff
@@ -63,10 +66,29 @@
...
+			{
+				userLocationAirport = userLocationObject.toString();
+			}
+			else
+			{
+				userLocationStatus = userLocationSetting.setPrimaryLocation(hsr, userId);
+				userLocationAirport = userLocationStatus.getPrimaryLocationAirport();
+			}
+			
+			if(request.getSearchSubmitFlag()==null && userLocationAirport!=null )
```
originalname51: I see the searchSubmitFlag is set in the flightcontroller.js as "Y". 

What is the code path where the value for search submit flag would be null?

gtas-parent/gtas-webapp/src/main/java/gov/gtas/controller/FlightPassengerController.java
```diff
@@ -63,10 +66,29 @@
 
 		String userId = GtasSecurityUtils.fetchLoggedInUserId();
 		boolean isAdmin = userService.isAdminUser(userId);
+		String userLocationAirport;
```
originalname51: I think it would be slightly more readable if the variables instantiated here were put closer to where they are actually used.

UserLocationAirport can probably go in the !Admin block, userlocationStatus can be assigned instead of created and then assigned, and default dest hashcode could be instantiated after the following if statement.

			if(request.getSearchSubmitFlag()==null && userLocationAirport!=null )

gtas-parent/gtas-webapp/src/main/java/gov/gtas/common/UserLocationSetting.java
```diff
@@ -37,73 +106,75 @@ public void setPrimaryLocation(HttpServletRequest httpServletRequest, String use
...
-				// if the user has a location but does not have a primary location, then set the
-				// first one as a primary location
-				if (!hasPrimaryLocation) {
+				/*
+				 * Case 1.2 
+				 *  The user has one or more locations but a primary location is not
+				 * enabled in the user_location table
+				 */
+
+				if (!userLocationStatus.isPrimaryLocationEnabledInDb()) {
```
originalname51: It looks like we are arbitrarily picking the last location a user has on their status if they do not have a primary location set in the database. Is there a way we can have more control over this? Would it make sense to treat this as an error and therefore present the dashboard airport instead?

gtas-parent/gtas-webapp/src/main/java/gov/gtas/common/UserLocationSetting.java
```diff
@@ -26,8 +26,77 @@
...
+			httpServletRequest.getSession().setAttribute(Constants.USER_PRIMARY_LOCATION,
+					userLocationStatus.getPrimaryLocationAirport());
+			logger.info(
+					"So user primary location exists so the user Location is just set in the HTTP session successfull");
+		}
+
+		else if (userLocationStatus.getPrimaryLocationAirport() != null && userLocationStatus.isPrimaryLocationCreated()
+				&& !userLocationStatus.isPrimaryLocationEnabledInDb()) {
+			/*
+			 * One or more location airports exist but none of them are marked primary in
```
originalname51: Commented on below

gtas-parent/gtas-webapp/src/main/java/gov/gtas/common/UserLocationSetting.java
kasimhelil: @originalname51 I discussed this with @TeddyTesfaye earlier today. If I've understood him correctly, `prime location` is the most recent location the user logged in.  So, if a user is associated with more than one location,  their prime location  changes as they change their login location.

gtas-parent/gtas-webapp/src/main/java/gov/gtas/common/UserLocationSetting.java
```diff
@@ -26,8 +26,77 @@
...
+				logger.info("One of the user's location airports is set as a primary location in user_location table");
+			} catch (Exception e) {
+				logger.error("An error has occurred when updating user location in the database.");
+			}
+
+		} else if (userLocationStatus.getPrimaryLocationAirport() != null
+				&& !userLocationStatus.isPrimaryLocationCreated()
+				&& !userLocationStatus.isPrimaryLocationEnabledInDb()) {
+			/*
+			 * No location airports were found in user_location table so insert the dashboard
```
originalname51: I like the auto-configure nature of the logic here but it seems like it would make more sense to do this upon user creation instead of on the fly.

gtas-parent/gtas-webapp/src/main/java/gov/gtas/common/UserLocationSetting.java
originalname51: I see userLocationStatus.setPrimaryLocationAirport setting the last item in the set as the primary location airport.

Issue #1264, opened 2019-07-31
Tool tips for Rules


originalname51:
On the rule page add help features so users can understand the types of rules they are creating.

An example of a help feature is a tooltip.

PR #1319, opened 2019-08-21
Tooltips for rules #1264

kasimhelil:
The `dictionary.json` file is not complete; more terms and their definitions are to be added to it when they are ready. The tooltip will display the term itself if its definition is not in the dictionary.

### DIFF
--- a/gtas-parent/gtas-webapp/src/main/webapp/build/BuildController.js
+++ b/gtas-parent/gtas-webapp/src/main/webapp/build/BuildController.js
@@ -3,7 +3,7 @@
  *
  * Please see LICENSE.txt for details.
  */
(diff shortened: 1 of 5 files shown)


gtas-parent/gtas-webapp/src/main/webapp/build/BuildController.js
```diff
@@ -521,8 +521,32 @@ app.controller('BuildController', function ($scope, $injector, jqueryQueryBuilde
...
+
+	// Adds tooltip to rule-value-container (for PNR's Trip Type)
+	$(document).on("mouseenter", "select[name='builder_rule_0_value_0']", function() {
+		$(this).find('option').each(function(){
+				$(this).prop('title', codeTooltipService.getCodeTooltipData($(this).text(), "dictionary"));
+		})
+		 
+	});
+
+	// Adds tooltip to rule-value-container (for Document Types and Passender Types)
```
originalname51: Are other entities like the flight handled here as well?

kasimhelil: Yes; they are handled.

PR #1343, opened 2019-08-29
Maven yarn

### LINKED ISSUE #1318: Maven Install not pulling latest NPM repo
Additions to the package.json file correctly cause the app to download new dependencies, but updates to the version number do not (always?) result in a new download. This is affecting deps installed with yarn.

gtasautumn:
Resolves #1318. Removed bower, upgraded dependencies to npm source repos, misc fixes, in order to correct the caching issue that prevented the build script from correctly fetching dependency versions when they were updated.


gtas-parent/gtas-webapp/src/main/webapp/app.js
```diff
@@ -526,11 +523,7 @@ var app;
                             };
                         },
                         passengers: function (paxService, $stateParams, paxModel) {
-                            //because of field/model not standard
-                            $stateParams.dest = $stateParams.destination;
-                            $stateParams.etaStart = $stateParams.eta;
-                            $stateParams.etaEnd = $stateParams.etd;
-                            return paxService.getPax($stateParams.id, paxModel.alldatamodel());
+                          return paxService.getPax($stateParams.id, paxModel.alldatamodel());
```
originalname51: Is this change a part of the update to npm?

originalname51:
Looks good! The web app project has another file, gulpfile.js located in the base directory of the gtas-webapps folder that looks like it could be in scope to remove here.

gtasautumn:
ha! Shouldn't you be watching fuzzy ultrasound images?

So yeah, interesting one. This is an undocumented fix for an issue that
fell out of that change. Problem was that the angular-ui-router bower
dependencies was a beta version and didn't have an exact version equivalent
in npm. I tested upgrading it to the nearest prod release in npm, which was
1.0.8 (ish), and *that* version caught a legitimate js exception that this
code causes.

Problem is that angular declares $stateParams non-extensible, so any code
that adds members to it is unambiguously misbehaving. Rather than roll the
dependency update back, I looked at the code in question and it is pretty
indefensible.

Further up the call stack is adding duplicate fields to $stateParams for
consumption here, apparently just so the field names match on assignment:

   someobject.dest = $stateParams.*dest*

Instead of writing:

   someobject.dest = $stateParams.*destination*

Where the data in the new *dest* field is identical to that in *destination*.
Same idea with 2 other date fields on $stateParams. It was cruft, so I
removed it, and no other issues came up with the new angular-ui-router
version.

Good eye, tho, and many many thanks for the attention you give our code. We
are going to miss you.

take care,
ab

Issue #1344, opened 2019-08-29
Flight Id Tag checks on views


TeddyTesfaye:
1. Added a condition under the where clause to check for flight_id_tag on both neo4j_vw and neo4j_hit_vw
2. Added a condition under the where clause to check for passenger_id_tag on the neo4j_hit_vw. Since the same check has already been implemented in neo4j_vw, the same check should be implemented in the neo4j_hit_vw so that it won't collect passengers that won't have id tags.
3. Since passenger_id_tag and flight_id_tag is used to identifiy a passenger and a flight in the node, other checks, a check on other passenger attributes and flight attributes is unnecessary so they are removed from the where condition

PR #1345, opened 2019-08-29
Added checks for flight id tag on the where close (of both views)

TeddyTesfaye:
@originalname51 
#1344 
Added checks for passenger id tag on the where close (for neo4j_hit_vw)
since there won't be any passengers without id_tag in neo4j view, the
corresponding hit_view should also check the for the same condition.

### DIFF
--- a/gtas-parent/scripts/db/1.2.1_to_1.6.0/neo4j_hit_vw.sql
+++ b/gtas-parent/scripts/db/1.2.1_to_1.6.0/neo4j_hit_vw.sql
@@ -1,4 +1,4 @@
-DROP VIEW IF EXISTS neo4j_vw;
+DROP VIEW IF EXISTS neo4j_hit_vw;
 CREATE VIEW neo4j_hit_vw AS
 SELECT
 	hs.id as gtas_hit_summary_id,
--- a/gtas-parent/scripts/db/1.2.1_to_1.6.0/neo4j_hit_vw.sql
+++ b/gtas-parent/scripts/db/1.2.1_to_1.6.0/neo4j_hit_vw.sql
@@ -1,0 +1,90 @@
+
+CREATE VIEW neo4j_hit_vw AS
+SELECT
+	hs.id as gtas_hit_summary_id,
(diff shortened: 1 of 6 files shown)


TeddyTesfaye:
@gtasautumn

originalname51:
Can you please add an update/create script as well?

TeddyTesfaye:
Added a rollback script

TeddyTesfaye:
Added create view script. FYI, It also exists in the gtas-neo4j-etl project

gtas-parent/scripts/db/1.2.1_to_1.6.0/neo4j_hit_vw.sql
```diff
@@ -0,0 +1,90 @@
+
+CREATE VIEW neo4j_hit_vw AS
```
originalname51: can we update this to create or replace? Likely scenario that this will be run is when the view will already exist.

gtas-parent/scripts/db/1.2.1_to_1.6.0/neo4j_vw.sql
```diff
@@ -0,0 +1,189 @@
+CREATE VIEW neo4j_vw AS
```
originalname51: can we update this to create or replace? Likely scenario that this will be run is when the view will already exist.

gtas-parent/scripts/db/1.2.1_to_1.6.0/neo4j_hit_vw.sql
```diff
@@ -0,0 +1,90 @@
+DROP VIEW IF EXISTS neo4j_vw;
```
originalname51: neo4j_vw is different than neo4j_hit_vw

TeddyTesfaye:
added a drop view as a first line (in case view exists)

originalname51:
looks good to me! While the work here is done in 1.7 the database changes are needed for 1.6 so I think the 1.2.1_to_1.6.0 is the best home for these scripts.

PR #1363, opened 2019-09-05
Trip type

gtasautumn:
Added the non-contiguous type to the enum and refactored the return value, added a chron sort on the leg number on the flightLeg entity, updated the trip type calculation for the non-contiguous type and refactored out extra for-loops, updated the front end to show a break for non-contiguous legs, and built a new leg array to correctly display leg endpoints for all trip types.


gtas-parent/gtas-commons/src/main/java/gov/gtas/enumtype/TripTypeEnum.java
```diff
@@ -31,29 +32,9 @@ public static TripTypeEnum getEnum(String value) {
...
-             returnString = "MULTI-CITY";
-           }
-           else if (this.name().equals("OPENJAW"))
-           {
-             returnString = "OPEN JAW";  
-           }
-           
-           return returnString;
-        }
+
```
originalname51: 👍

gtas-parent/gtas-commons/src/main/java/gov/gtas/model/FlightLeg.java
```diff
@@ -56,5 +56,13 @@ public BookingDetail getBookingDetail() {
 
 	public void setBookingDetail(BookingDetail bookingDetail) {
 		this.bookingDetail = bookingDetail;
-	}
+  }
+  
+  @Override
+  public int compareTo(FlightLeg leg) {
+    if (getLegNumber() == null || leg.getLegNumber() == null) {
```
originalname51: Flight legnumber is defined as non-null. Is this null check required?
    @Column(name = "leg_number", nullable = false)
    private Integer legNumber;

The compareTo documentation does suggest that because this will not be the same as equals to add the following comment on the class:
"Note: this class has a natural ordering that is inconsistent with equals."

gtas-parent/gtas-loader/src/main/java/gov/gtas/services/PnrMessageService.java
```diff
@@ -238,71 +238,70 @@ protected void updatePaxEmbarkDebark(Pnr pnr) {
...
+          noncontig = true;
+        }
+        previousDest = getFlightLegDest(flightLeg);
+      }
+
+        // check dwell times for ones over 24 hours
+        for (DwellTime dwellTime : dwellTimeSet) {
+          if (dwellTime.getDwellTime() != null) {
+              double dwellTimeHours = dwellTime.getDwellTime();
+              if (dwellTimeHours > 24.0) {
```
originalname51: Out of scope but I wonder if making the long dwell time configurable would be useful.

gtasautumn: glad you reminded me of that. Been meaning to ask someone how we arrived at 24 hours and whether that was written in stone or could vary. Either way, I think this code could use some refactoring. thx.

Issue #1322, opened 2019-08-22
On user creation a blank value is populated on flight filter instead of user location


originalname51:
Steps to reproduce:

1. make new user
2. log in to gtas
3. open flight page.
4. notice the location is blank.

Issue #1320, opened 2019-08-21
User Location: Add a feature to auto-populate Origin and Destination when changing direction


TeddyTesfaye:
Add a feature to auto-populate Origin and Destination when changing direction when selecting inbound/outbound from direction drop down.

PR #1364, opened 2019-09-05
Dev

TeddyTesfaye:
@kasimhelil 

#1322 
#1320


originalname51:
The utilcontroller class will attempt to load an application context and will fail if the database is not running / able to be connected to. 

The project will look for classes appended or prepended with "Test" when running unit test (which mvn clean install is defaulted to do). 

Please rename UtilControllerTest to UtilControllerIT as it requires dependencies to be loaded.

gtas-parent/gtas-webapp/src/test/java/gov/gtas/controller/UtilControllerTest.java
```diff
@@ -0,0 +1,95 @@
...
+import gov.gtas.common.UserLocationSetting;
+import gov.gtas.common.UserLocationStatus;
+import gov.gtas.common.WebAppConfig;
+import gov.gtas.controller.config.TestMvcRestServiceWebConfig;
+import gov.gtas.services.security.UserService;
+
+@RunWith(SpringJUnit4ClassRunner.class)
+@ContextConfiguration(classes = { TestMvcRestServiceWebConfig.class, WebAppConfig.class })
+@WebAppConfiguration
+public class UtilControllerTest {
```
originalname51: Please rename to UtilControllerIT

originalname51:
Looks good to me

originalname51:
I will add a unit test.

PR #1369, opened 2019-09-07
Improved performance on loading passengers after clicking on a flight…

originalname51:
…. There was a call in a loop to the database to get seat information. This is not necessary for populating the grid and was taking up about 2.1 seconds.

The majority of the time to load the page now comes from the getPassengerByCritera method.

### DIFF
--- a/gtas-parent/gtas-commons/src/main/java/gov/gtas/model/Document.java
+++ b/gtas-parent/gtas-commons/src/main/java/gov/gtas/model/Document.java
@@ -39,8 +39,12 @@ public class Document extends BaseEntity {
 	@Column(name = "issuance_country")
 	private String issuanceCountry;
 
+	@JoinColumn(name = "passenger_id", columnDefinition = "bigint unsigned")
 	@ManyToOne(fetch = FetchType.LAZY)
 	private Passenger passenger;
+
+	@Column(name = "passenger_id", columnDefinition = "bigint unsigned", insertable = false, updatable = false)
+	private Long paxId;
 
     /** calculated field */
     @Column(name = "days_valid")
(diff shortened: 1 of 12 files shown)


gtas-parent/gtas-commons/src/main/java/gov/gtas/services/PassengerServiceImpl.java
```diff
@@ -137,15 +137,6 @@ public PassengersPageDto getPassengersByCriteria(Long flightId, PassengersReques
...
     		}
-            
-            List<Seat> seatList = seatRepository.findByFlightIdAndPassengerId(f.getId(), p.getId());
-            if (CollectionUtils.isNotEmpty(seatList)) {
-                List<String> seats = seatList.stream().map(seat -> seat.getNumber()).distinct()
-                        .collect(Collectors.toList());
-                if (seats.size() == 1) {
-                    vo.setSeat(seats.get(0));
-                }
-            }
```
tsegay: The passenger vo returned from the service will always have the seat field `null`. This will cause issues when the the sear number is expected in the frontend. 

As per our internal discussion, you can ...

1. Refactor the code to avoid the database trip to get the seat number

2. Introduce a new VO for the datagrid that does not have a seat number.

originalname51: Updated with a new VO, I also did #1372 as it was in the same code-path.

tsegay: Looks good.

gtas-parent/gtas-commons/src/main/java/gov/gtas/model/Document.java
```diff
@@ -117,6 +128,7 @@ public boolean equals(Object obj) {
 		final Document other = (Document) obj;
 		return Objects.equals(this.documentNumber, other.documentNumber)
 				&& Objects.equals(this.issuanceCountry, other.issuanceCountry)
-				&& Objects.equals(this.documentType, other.documentType);
+				&& Objects.equals(this.documentType, other.documentType)
+				&& Objects.equals(this.paxId, other.paxId);
```
tsegay: please add unit tests for `hashCode() `and `equals(Object obj)`.

originalname51: Updated unit test

PR #1375, opened 2019-09-10
978

### LINKED ISSUE #978: Intra-Flight Entity Resolution
using reservation system provided unique IDs to resolve passengers within a "family" of PNR messages.

kasimhelil:
Resolves #978


gtas-parent/gtas-commons/src/test/java/gov/gtas/repository/FlightPassengerRepositoryTest.java
```diff
@@ -0,0 +1,39 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.junit.MockitoJUnitRunner;
+
+import gov.gtas.model.FlightPassenger;
+
+@RunWith(MockitoJUnitRunner.class)
+public class FlightPassengerRepositoryTest {
```
originalname51: I think this test class isn't testing the logic of the application. 

repository is mocked and so TestsGetPassengerUsingREF5 will never fail. 


I think we should removing this test and focusing on testing the services and other application logic.

gtas-parent/gtas-loader/src/main/java/gov/gtas/services/ServiceUtil.java
```diff
@@ -55,9 +55,19 @@ public Passenger findPassengerOnFlight(Flight f, PassengerVo pvo) {
         if (f.getId() == null) {
             return null;
         }
+        List<FlightPassenger> pax = null;
+        
+        //Query passenger by Flight -> pnr recordLocator# -> pnr REF number
```
originalname51: The REF field is conditional so it is possible to have a passenger who comes in initially without this information, and then has the information later assigned. I think instead of an else statement we should have logic that executes when pvo.getPnrRecordLocator() != null && pvo.getPnrReservationReferenceNumber() != null and the results of the aforementioned also result in no passenger returned.

gtas-parent/gtas-loader/src/test/java/gov/gtas/services/ServiceUtilTest.java
```diff
@@ -0,0 +1,52 @@
...
+	
+	@Mock
+	Passenger passenger;
+	
+	@Mock
+	Passenger passenger2;
+	
+	@Test
+	public void testFindPassengerOnFlight() {
+		Mockito.when(service.findPassengerOnFlight(flight, pvo)).thenReturn(passenger);
```
originalname51: This is testing a mock and should test results of the service instead.

gtas-parent/gtas-loader/src/main/java/gov/gtas/services/ServiceUtil.java
kasimhelil: I think having the else block here is fine; otherwise, we will query twice. Or we can check if `pax` is not null  and run the code inside the `else` block which will do the same thing.
**Note:** When we decide on the entity resolution staff, we should change the code inside the else block.

gtas-parent/gtas-commons/src/test/java/gov/gtas/repository/FlightPassengerRepositoryTest.java
kasimhelil: I can't think of any other way to unit test this🤦🏼‍♂️I will remove it!!

gtas-parent/gtas-commons/src/main/java/gov/gtas/repository/FlightPassengerRepository.java
```diff
@@ -23,4 +24,23 @@
...
+             "LEFT JOIN FETCH pax.seatAssignments " +
+             "LEFT JOIN FETCH pax.bags " +
+             "LEFT JOIN FETCH pax.flightPaxList " +
+             "LEFT JOIN FETCH pax.tickets " +
+             "LEFT JOIN FETCH pax.bookingDetails " +
+             "LEFT JOIN FETCH pax.passengerDetails " +
+             "LEFT JOIN FETCH pax.passengerTripDetails " +
+             "WHERE fp.flightId = :flightId " +
+             "AND :recordLocator IN (SELECT pnr.recordLocator FROM fp.flight.pnrs pnr) " +
+             "AND fp.passenger.passengerTripDetails.pnrReservationReferenceNumber = :pnrReservationReferenceNumber ")
```
originalname51: please name fp.passenger. to pax

gtas-parent/gtas-loader/src/main/java/gov/gtas/services/ServiceUtil.java
tjwdev: Querying more than once is an unavoidable consequence of our attempt to prove the negative. We could cache the passenger list we're querying against, but this will be done on a small enough set that the impact should be low enough as to be negligible.

What needs to be decided on before this ticket is complete/merge-able is what we should do our secondary check on  within the else (most likely the hash) and what we should do with the hash on personal details that change over time.

Issue #1174, opened 2019-06-04
Manage Timezones


originalname51:
AC1: Convert time to UTC in database
AC2: Store any values coming in that are not UTC in their own column/table (keep UTC conversion)
AC3: Front end changes will be in local time of the airport (e.g. if GTAS is loaded in EST timezone all flights are EST).

originalname51:
Flight leg highlighting might need to be tweaked.

originalname51:
No tweaks required - highlighting is done on origin/dest.

dje04001:
Passenger Details > PNR > Itinerary table, ETA is off by 6 hours on a flight from AMS>IAD

dje04001:
Looks good now with the latest fixes.

PR #1377, opened 2019-09-11
UTC Time Changes - #1174 - Updated loading to put in UTC time for fli…

### LINKED ISSUE #1376: Flight eta/etd not shown on flight history
Flight eta/etd not shown on flight history  - this is due to not having a mutable flight details object copied into the flight history.

originalname51:
…ghts.

Several files changed, all typically relating to local etd/eta date and utc etd/eta.

Also fixed #1376 as it was in codepath.


gtas-parent/gtas-loader/src/main/java/gov/gtas/services/GtasLoaderImpl.java
```diff
@@ -267,14 +270,20 @@ public Flight processFlightsAndBookingDetails(List<FlightVo> flights, Set<Flight
...
                   // Special case where prime flight doesnt have a timestamp
                     //  we use the prime TDT with the 125 or 87 LOC code \ 189 DTM code for APIS
                     //  or tvl level 0 etd timestamp for PNR if this is the case.
-                    mfd.setEtd(primeFlightTimeStamp);
+                    mfd.setLocalEtdDate(primeFlightTimeStamp);
                 }
+                String originAirport = currentFlight.getOrigin();
+                String destinationAirport = currentFlight.getDestination();
+                Date utcETADate = gtasLocalToUTCService.convertFromAirportCode(destinationAirport, fvo.getLocalEtaDate());
+                Date utcETDDate = gtasLocalToUTCService.convertFromAirportCode(originAirport, fvo.getLocalEtdDate());
```
gtasautumn: Is the ETA/ETD data in a single message local to the arrival/destination *airports* or to the message receiver location??? I thought all non-UTC datetime fields were local to the receiver, which would mean we should be converting data to UTC based on the receiver location only, not to both flight endpoints.

So if I am at Townsville USA International, receiving messages for flight 0100 from Spain to Townsville, that data is meant for my location so all departure and arrival datetimes would be in my local time, correct?

originalname51: From my understanding it's the local time of the airport it is coming from and so we would have to account for USA, Spain, and UTC in the example given above. The spec docs are not super clear here - @dje04001 is this correct?

PR #1378, opened 2019-09-11
Added Code/unit tests to display application version number.

TeddyTesfaye:
@tsegay 
@kasimhelil 
@gtasautumn 

Please review and merge.

Thanks,
Teddy

### DIFF
--- a/gtas-parent/gtas-webapp/src/main/java/gov/gtas/controller/VersionController.java
+++ b/gtas-parent/gtas-webapp/src/main/java/gov/gtas/controller/VersionController.java
@@ -1,0 +1,31 @@
+package gov.gtas.controller;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.web.bind.annotation.RequestMapping;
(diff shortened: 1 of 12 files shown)


gtas-parent/gtas-webapp/src/main/java/gov/gtas/controller/VersionController.java
```diff
@@ -0,0 +1,31 @@
...
+
+import gov.gtas.services.VersionService;
+
+@RestController
+public class VersionController {
+
+	private static final Logger logger = LoggerFactory.getLogger(VersionController.class);
+
+	@Autowired
+	private VersionService versionService;
```
originalname51: Field injection is not recommended by spring. Injecting through a constructor is recommended. 

From the spring documents: 
[
The Spring team generally advocates constructor injection, as it lets you implement application components as immutable objects and ensures that required dependencies are not null](https://docs.spring.io/spring/docs/current/spring-framework-reference/core.html)

TeddyTesfaye:
Good point on the documentation, required vs optional dependencies. Since we have both on the code base and this task is done, I will consider constructor dependency on the upcoming changes.
Thanks,
Teddy

originalname51:
Looks solid

PR #1381, opened 2019-09-12
Message persistence

### LINKED ISSUE #1312: Create Cache For Incoming Messages
Create a cache for incoming messages. 

Write all incoming messages to the cache.

Create a task to read and process cache on bean creation

Remove messages from cache upon finishing the file.

This will allow for no data loss on server shutdown.

gtasautumn:
Resolves #1312

Fix is to persist the message files to disk at all times so they can't be lost in the event of a catastrophic failure during processing. **Two new folders must be added for files in-process or which failed, both need references in application.properties**. The default folders are MESSAGEWORKING and MESSAGEERROR in the same parent dir as the MESSAGE and MESSAGEOLD (incoming and outgoing) folders.

Messages are now moved from MESSAGE to MESSAGEWORKING as each file is read from the queue into memory. The worker threads operate on the files in MESSAGEWORKING and place them in MESSAGEOLD when done. Files are renamed with a timestamp suffix if the target dir already has a file with the same name, and files are no longer deleted or orverwritten for any reason. **Files are moved using the StandardCopyOption.ATOMIC_MOVE option, which has only been tested on windows so far**.

MESSAGEERROR for now just has the files that trigger exceptions during the move process (IOExceptions from missing directories, permissions issues, etc), but there will be another ticket to scrub the file processing code to see what other errors should drop files there.


gtas-parent/gtas-job-scheduler-war/src/main/java/gov/gtas/job/scheduler/LoaderScheduler.java
```diff
@@ -118,23 +121,43 @@ private void processSingleFile(File f, LoaderStatistics stats, String[] primeFli
...
+    boolean saved = false;
+
+    if (file == null || !file.isFile()) return saved;
+
+    FileSystem toFileSystem = FileSystems.getDefault();
+    String fileName = file.getName();
+
+    try {
+      Path destPath = toFileSystem.getPath(messageProcessedDir + File.separator + fileName);
+      logger.info("Moving file to PROCESSED: " + destPath.toString());
```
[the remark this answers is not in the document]
originalname51: This will exacerbate the logging issue as we do not need to print the file name to the logs as we should be able to find through the database. We can monitor with a performance test to see if this would be detrimental to what information logging prints under load.

gtas-parent/gtas-job-scheduler-war/src/main/java/gov/gtas/job/scheduler/Utils.java
```diff
@@ -0,0 +1,118 @@
...
+    String text = null;
+    if (!exists(filePath)) return text;
+
+    try {
+      byte[] raw = FileUtils.readSmallFile(filePath);
+      String tmp = new String(raw, StandardCharsets.UTF_8);
+      text = ParseUtils.stripStxEtxHeaderAndFooter(tmp);
+      }
+      catch (IOException ex) {
+        //??
```
originalname51: even if we eat and ignore an exception we could set the exception name to 'ignored' and log it.

gtas-parent/gtas-job-scheduler-war/src/main/java/gov/gtas/job/scheduler/LocalFileIntakeScheduler.java
```diff
@@ -33,8 +36,11 @@
     @Value("${message.dir.origin}")
     private String messageOriginDir;
 
-    @Value("${message.dir.processed}")
-    private String messageProcessedDir;
+    @Value("${message.dir.working}")
```
originalname51: These are not used in this file.

gtas-parent/gtas-job-scheduler-war/src/main/java/gov/gtas/job/scheduler/LoaderWorkerThread.java
```diff
@@ -57,14 +56,10 @@ public void run() {
...
+
 					MessageHeaders headers = msg.getHeaders();
-					String filename = (String) headers.get("filename");
-					if (filename != null) {
-						this.fileName = filename;
-					} else {
-						this.fileName = UUID.randomUUID().toString();
-					}
-					this.text = msg.getPayload().toString();
+          fileName = (String) headers.get("filename");
```
originalname51: Filename can be null here. When  loader.receiveMessage is called in loader worker thread it uses this to create a file.

    File f = new File(messageWorkingDir + File.separator + fileName);

originalname51: Nevermind! I see this is handled in the initial jms loader!

PR #1408, opened 2019-09-19
fix to validate required fields

### LINKED ISSUE #1405: Create Case - saving null data causes error
Data should be validated before submit for required fields.

gtasautumn:
simple notification of missing data and will not submit the form without it. resolves #1405.


kasimhelil:
It looks good!

PR #1431, opened 2019-09-27
Translation 1410

gtasautumn:
code for #1429 #1412 #1427, includes the flight, flightpax, paxdetails, cases, and 1 day lookout pages.

All text (exceptions described at the end) displayed on these pages should reflect the data in the message_lang.properties files, and the language displayed is pulled from the browser default parent language, ie for `en-US`, it will choose the language file for `en`.

I have updated the language files for `en` and `es` only for testing. The terms in the `es` file are incomplete, but where we don't yet have a Spanish translation I have included the English one with a `^` sign at the end.

in Chrome, 

1. view the pages in English and verify that all text displays as expected and does not include the `^` character. 
2. Under Chrome setttings, add Spanish, select the option to display Chrome in Spanish, and move Spanish to the top of your language list.

image

3. Click Reiniciar (reload) Chrome

4. In the app, logout and log back in
5. Review the pages in Spanish (and some English with `^`) and verify the text displays as expected.

**EXCEPTIONS TO TRANSLATED TEXT**
No grid data is currently translated or any hover text for grid data, and no text in the footer or hamburger menu in the grids. These will still be in English:

image

image

image

PR #1436, opened 2019-10-02
Translation 1435

gtasautumn:
Affects the title and query-limit notification header on the query flights page.

To test, build any query in query builder and click the search flights button
   image

Then verify that the title banner on the flights page shows the translated text (see #1431 for instructions). To force the query-limit banner, you can cheat and update the flag in the code in flightscontroller.js. Set the value for `$scope.queryLimitReached`  to `true`  on line 106.
   image


kasimhelil:
I think the destination branch for the pull request should be `dev`, instead of `master`.  Would you please change that and make another pull request?

gtasautumn:
changed the target branch to dev.

Issue #1447, opened 2019-10-08
Develop Message Delivery Compliance Report using Kibana Canvas


TeddyTesfaye:
Related to #1299
Completed

Issue #1446, opened 2019-10-08
Indexing Data in ES for Reporting


TeddyTesfaye:
Related to #1299

Completed.

Issue #1299, opened 2019-08-06
Research Indexing of Message Data for reporting and Kibana Canvas


dje04001:
Partners require the ability to monitor message flow by carrier, by flight, by push, and by day through a report. It appears that we could do this fairly easily in Kibana Canvas if we indexed message data.

1. Is this possible?
2. What is the level of effort required to index this data?
3. Can we use Canvas to fulfill this requirement?

TeddyTesfaye:
Completed

PR #1449, opened 2019-10-08
Added Scripts, and configurations for Message Count and Message Delivery

TeddyTesfaye:
Related to #1299, #1446, #1447, # 1448

@tsegay @kasimhelil


gtas-parent/gtas-commons/src/main/resources/sql/message_delivery_comp_vw.sql
```diff
@@ -0,0 +1,63 @@
+CREATE VIEW message_delivery_comp_vw AS
```
originalname51: To prevent errors on running hibernate create multiple times i suggest updating from create view to 

CREATE OR REPLACE VIEW

gtas-parent/scripts/elastic/config/logstash/message_delivery_comp_script.sql
```diff
@@ -0,0 +1 @@
+select * from message_delivery_comp_vw
```
originalname51: We might want to consider a limit on this query in order to batch process

tsegay: We don't need to limit the query, logstash will handle the paging with the  following configuration.

```
    jdbc_paging_enabled => "true"
    jdbc_fetch_size => "10000"
```

TeddyTesfaye: Yeap, that is handled in logstash so no need to do it

originalname51: cool!

Issue #1468, opened 2019-10-28
PnrGovParserTest: Failure in different timezone


yashpatil:
When running a `mvn clean install` in a different timezone, the PnrGovParser `pnrDocoTest` fails while asserting the `doco issuance date` as the `docoVo.getIssuanceDate().toString()` method will format it for the local timezone. e.g.

```
org.junit.ComparisonFailure: 
Expected :Fri Dec 22 00:00:00 NZDT 2006
Actual   :Fri Dec 22 00:00:00 EST 2006
```

Referring to the EDIFACT [specifications](https://www.icao.int/Security/FAL/Documents/1-PNRGOV-EDIFACT_Implementation-Guide-13-1version-SECOND.pdf), this section contains document information for identification purposes, which would most probably omit the time information in the issuance date section:

```
A passenger by the name of Mr. John Meeks supplies a United States Redress number for his PNR:
a. For those systems using automated format:
SSR+DOCO:HK:1:AA:::JFK:LAX:0001Y28JUN//R/1234567890123///US
b. For those systems using non-automated format:
SSR+DOCO:HK:1:AA::::://R/1234567890123///US
```

Based on the above, the comparison test should not use the time component of the date.

originalname51:
Code in dev; Yashpatil completed, I reviewed.

PR #1469, opened 2019-10-28
1468 - PnrGovParserTest fix failure in different time zone

yashpatil:
Fix for [1468](https://github.com/US-CBP/GTAS/issues/1468).

This issue fixes a test failure for the `PnrGovParserTest` when run in a different time zone.

The `pnrDocoTest()` should ignore the time component of `doco issuance date` during test assertion.


originalname51:
Good catch!

PR #1472, opened 2019-10-28
Failed email handling

simbamarufu1:
Should an email fail to send, this failure should not cause the entire batch to fail.


gtas-parent/gtas-commons/src/main/java/gov/gtas/services/NotificatonServiceImpl.java
```diff
@@ -121,12 +120,16 @@ public NotificatonServiceImpl(SnsService snsService,
 
 	@Override
 	@Transactional
```
simbamarufu1: This doesn't seem needed if the class has `@Transactional`

originalname51: We should look into removing this from the class or method.

Issue #1476, opened 2019-10-29
Integrate PDF report with with application


originalname51:
Implement new PDF report with web application

Issue #1475, opened 2019-10-29
Implement new pdf report


originalname51:
Implement a new PDF report, taking into account all information from flight details page

Issue #1473, opened 2019-10-29
Integrate new PDF library with GTAS


originalname51:
Keep in mind licensing restrictions

PR #1499, opened 2019-11-04
Dev

TeddyTesfaye:
#1473, #1475, #1476 
@tjwdev 
@tsegay 
@kasimhelil 
@originalname51


gtas-parent/checkstyle.xml
```diff
@@ -1,10 +0,0 @@
-<!DOCTYPE module PUBLIC
```
originalname51: Was this causing issues? It does not appear to be related to other code but is deleted.

TeddyTesfaye: No it did not. Hmm.. Let me see why it is deleted

Issue #1512, opened 2019-11-06
Paxdetails - fix translations on PNR state and Baggage pax name


gtasautumn:
The translation code shows up for PNR Address grid where column = state:

![uploads/febdc604-d501-4c79-b561-5f74863ec421/image.png](https://s3-us-west-2.amazonaws.com/prod.huboard.com/uploads%2Ffebdc604-d501-4c79-b561-5f74863ec421%2Fimage.png)

and the column header under Total Baggage says 'Passenger Last Name' when the full name is displayed:

![uploads/46cb99bb-9e6e-4a28-9808-9055bada800e/image.png](https://s3-us-west-2.amazonaws.com/prod.huboard.com/uploads%2F46cb99bb-9e6e-4a28-9808-9055bada800e%2Fimage.png)

originalname51:
Validated in dev environment

PR #1513, opened 2019-11-06
fixed the address state translation code, renamed the Baggage Pax Nam…

gtasautumn:
…e column

Fixed in the English and Spanish test language files.

The Total PNR Baggage grid column was called Pax Last Name, but returned the pax full name. Changed this column header to just read 'Passenger'.

Also fix the code mapping for the Address grid's State col so it reads "State/Province" to better suit international addresses.


gtasautumn:
See #1512 for images and more notes

PR #1544, opened 2019-11-21
remove 1day lookout role from script

gtasautumn:
#1532


gtasautumn:
> Looks good. Eventually we may want to revisit the naming conventions of "Manage Cases".

Agreed

PR #1566, opened 2019-11-25
UTF-8 4 byte character set and log fix

simbamarufu1:


tsegay:
Looks like this is a PR against master ?

PR #1569, opened 2019-11-27
Using docker for deployments

simbamarufu1:


docker-compose-original.yml
```diff
@@ -0,0 +1,230 @@
+version: "3.4"
```
simbamarufu1: This is the existing docker-compose file. I am keeping it here as we add more docker-compose files for each set of services

docker-compose.yml
```diff
@@ -1,230 +1,56 @@
...
+  mysql_processor_password:
+    file: ./mariadb/mysql_processor_password.txt
+volumes:
+  gtas-parent:
+    driver_opts:
+      type: none
+      device: ${PWD}/gtas-parent
+      o: bind
+  db-data:
+  m2:
```
simbamarufu1: Driver defaults to local, but will need to do research for performance tuning so docker-compose file can be used for both production and local testing. Same applies to the parent volume

docker-compose.yml
```diff
@@ -1,230 +1,56 @@
...
+    secrets:
+     - mysql_root_password
+    ports:
+      - 3306:3306
+    networks:
+      - gtas_network
+    restart: unless-stopped
+    volumes:
+      - db-data:/var/lib/mysql
+secrets:
```
simbamarufu1: Does anyone know how to add all the secrets in one file?

mariadb/init/mariadb-user-creation-script.sh
```diff
@@ -0,0 +1,29 @@
+host=localhost
```
simbamarufu1: Will add more init files

gtas-parent/gtas-commons/pom.xml
```diff
@@ -383,10 +383,6 @@
 						<version>1.1.1</version>
 					</dependency>
 				</dependencies>
-				<configuration>
```
simbamarufu1: This property overrides the hibernate.properties url, while there is a warning about not removing this here:
http://juplo.de/hibernate4-maven-plugin/configuration.html 

It seems safe to remove, but I would appreciate more eyes on this

tsegay:
### Test Scenario: 

Given the [Dockerfile](https://github.com/US-CBP/GTAS/blob/fd5792f7bafd8e664354add8d434fe8849f850dc/gtas-parent/Dockerfile) AND local database running on docker (hostname = mariahost, network = es_gtas_default)

**Build a docker image from the commandline**
(Inside gtas-parent)$ `docker build -t gtaswco/webapp .`

**Once the image is created, launch the webapp container**  
(Inside gtas-parent)$ `docker run --rm -it --name wcogtas -p 8080:8080 --network es_gtas_default -e "DB_HOST=mariahost" -e "DB_USERNAME=root" -e "DB_PASSWORD=admin"  gtaswco/webapp:latest`

### Problem: 
The environment variables, I'm passing in, wont override the values in the application.properties because currently, it is hardcoded. 

### Suggestions for a fix:
1. Replace configurable values with environment variables (database host, username, password) then provide a way to pass in new values when building an image or running the container.
1. (When building the image) Define ARG and ENV values in the Dockerfile. You can have default values set there and have an option to override that when building the docker image. 

      (Inside gtas-parent)$ `docker build --build-arg DB_HOST=mariahost --build-arg DB_USERNAME=root --build-arg DB_PASSWORD=admin -t gtaswco/webapp .`

2. (When running the container) You can provide values for the environment variables.

     (Inside gtas-parent)$ `docker run --rm -it --name wcogtas -p 8080:8080 --network es_gtas_default -e "DB_HOST=mariahost" -e "DB_USERNAME=root" -e "DB_PASSWORD=admin"  gtaswco/webapp:latest`

PR #1574, opened 2019-12-06
Update issue templates

dje04001:


tsegay:
Creates a default bug template and a default configuration file for the bug template

PR #1589, opened 2019-12-17
Fixed passenger ref compare & prefix/suffix bug in parser

tjwdev:
-modified ref number comparison in PnrGovParser to appropriately use the
TIF traveler details reference number, it was using TRI erroneously.

-modified the tif name back up comparison by appropriately removing
suffixes and prefixes from the first and last names before being
compared, this involved adding two new utility methods to the PnrUtils
class.

Issue wound up being parsing error rather than post-parse error.


gtas-parent/gtas-parsers/src/main/java/gov/gtas/parsers/pnrgov/PnrGovParser.java
```diff
@@ -652,21 +652,26 @@ private void processGroup6_Agent(DAT_G6 dat, TVL tvl, FlightVo flightVo) throws
...
+		if(tif != null) {
 			List<TravelerDetails> td = tif.getTravelerDetails();
-			if (CollectionUtils.isNotEmpty(td)) {
-				String firstName = td.get(0).getTravelerGivenName();
-				for (PassengerVo pax : parsedMessage.getPassengers()) {
-					if (surname.equals(pax.getLastName()) && firstName.equals(pax.getFirstName())) {
-						currentPassenger = pax;
-						break;
+			if(CollectionUtils.isNotEmpty(td)) {	
+				refNumber = tif.getTravelerDetails().get(0).getTravelerReferenceNumber(); //Grab TIF Ref number, which is number by which passengers are ordered on the PNR
```
originalname51: can 
tif.getTravelerDetails() 

or
 tif.getTravelerDetails().get(0)

ever result in a null value?

originalname51:
looks good; I think a unit test would be really helpful and doublechecking if null can be returned on the tif.getTravelerDetails() or tif.getTravelerDetails().get(0)

gtas-parent/gtas-parsers/src/main/java/gov/gtas/parsers/pnrgov/PnrGovParser.java
tjwdev: They can result in nulls, but both are checked before attempting to use. isNotEmpty() is a null-safe check.

PR #1671, opened 2020-01-13
Neo4J, ETL, scheduler, and active mq

simbamarufu1:


neo4j-etl-docker-compose.yml
```diff
@@ -0,0 +1,93 @@
...
+    environment:
+      DB_HOST: mariadb
+      KIBANA_HOST: kibana
+      NEO4J_HOST: neo4j
+      ACTIVEMQ_HOST: activemq
+    volumes:
+      - type: bind
+        source: ## your local path to gtas_documents
+        target: /usr/local/gtas_documents
+        consistency: consistent
```
simbamarufu1: We are reliant on this synchronization process being fast, we will need to do some performance testing

gtas-parent/docker-resources/default.application.properties
```diff
@@ -59,10 +59,10 @@ hibernate.dialect=org.hibernate.dialect.MySQL5InnoDBDialect
 allow.loosen.resolution=true
 inputType=two_dirs
 
-message.dir.origin=C:\\MESSAGE
-message.dir.working=C:\\MESSAGEWORKING
-message.dir.processed=C:\\MESSAGEOLD
-message.dir.error=C:\\MESSAGEERROR
+message.dir.origin=/usr/local/gtas_documents/input
+message.dir.working=/usr/local/gtas_documents/working
+message.dir.processed=/usr/local/gtas_documents/processed
```
simbamarufu1: Is it necessary to have `/processed` mounted on a volume?

tsegay: I would say Yes. Otherwise, it will be hard to clean up files when you run out of space.

originalname51: second the importance; it is important that the files are accessible outside of gtas for maintenance and troubleshooting.

Issue #1730, opened 2020-02-24
Add endpoint on loader for messages from web


originalname51:
Add a message endpoint to the loader so messages can be posted to GTAS.

PR #1731, opened 2020-02-24
#1730 Adding an endpoint for messages to be received into GTAS. At th…

originalname51:
…is point this is primarily for testing. Once security is implemented this will be another way airlines can communicate APIS/PNR to GTAS.

### DIFF
--- a/gtas-parent/docker-resources/default.application.properties
+++ b/gtas-parent/docker-resources/default.application.properties
@@ -55,7 +55,9 @@ hibernate.connection.charSet=utf8
 hibernate.dialect=org.hibernate.dialect.MySQL5InnoDBDialect
 
 ##################### MessageLoader #####################
-
+#the scheduler.web.receiver.on is an unsecured endpoint that accepts messages and puts them on the internal queue.
+# There is *no* security on the endpoint and so unless there is outside security provided it should not be turned on.
+scheduler.web.receiver.on = false
 allow.loosen.resolution=true
 inputType=two_dirs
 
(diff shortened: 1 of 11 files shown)


gtas-parent/docker-resources/default.application.properties
```diff
@@ -55,7 +55,9 @@ hibernate.connection.charSet=utf8
 hibernate.dialect=org.hibernate.dialect.MySQL5InnoDBDialect
 
 ##################### MessageLoader #####################
-
+#the scheduler.web.receiver.on is an unsecured endpoint that accepts messages and puts them on the internal queue.
+# There is *no* security on the endpoint and so unless there is outside security provided it should not be turned on.
+scheduler.web.receiver.on = false
```
simbamarufu1: 👍

gtas-parent/gtas-job-scheduler-war/src/main/java/gov/gtas/job/scheduler/controller/MessageReceiverController.java
```diff
@@ -0,0 +1,42 @@
...
+
+	MessageReceiverService messageReceiverService;
+
+	public MessageReceiverController(MessageReceiverService messageReceiverService) {
+		this.messageReceiverService = messageReceiverService;
+	}
+
+	@PostMapping(value = "/api/message", consumes = MediaType.APPLICATION_JSON_VALUE)
+	public void consumeMessage(@RequestBody WebMessage messagePayload) {
+		if (messagePayload == null || messagePayload.getMessagePayload() == null) {
```
simbamarufu1: We have this same check in the service but handled differently. Is that intentional?

originalname51: Yes, perhaps I was over-thinking the message receiver service being used elsewhere, and I wanted to log any messages that were null.

PR #1736, opened 2020-03-01
1536 1735 bags and hours before

originalname51:
This code will allow running rules against null bags by making null bag = no bag.

This behavior is toggle-able.

### DIFF
--- a/gtas-parent/gtas-loader/src/main/java/gov/gtas/services/GtasLoaderImpl.java
+++ b/gtas-parent/gtas-loader/src/main/java/gov/gtas/services/GtasLoaderImpl.java
@@ -10,6 +10,7 @@ import gov.gtas.parsers.exception.ParseException;
 import gov.gtas.parsers.util.DateUtils;
 import gov.gtas.parsers.vo.*;
 import gov.gtas.repository.*;
+import gov.gtas.vo.passenger.CountDownVo;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.BeanUtils;
@@ -17,6 +18,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
 import java.util.*;
+import java.util.concurrent.TimeUnit;
 
 import static gov.gtas.services.LoaderUtils.getNullPropertyNames;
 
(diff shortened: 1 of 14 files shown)


gtas-parent/gtas-loader/src/main/java/gov/gtas/services/GtasLoaderImpl.java
```diff
@@ -333,12 +335,46 @@ public PassengerInformationDTO makeNewPassengerObjects(Flight primeFlight, List<
...
+			Date flightDate = primeFlight.getMutableFlightDetails().getEtd();
+			CountDownCalculator countDownCalculator = new CountDownCalculator(transmissionDate);
+			CountDownVo countDownVo = countDownCalculator.getCountDownFromDate(flightDate);
+			if (countDownVo.getMillisecondsFromDate() > 0
+					&& countDownVo.getMillisecondsFromDate() <= Integer.MAX_VALUE) {
+				hoursBeforeTakeOff = (int) TimeUnit.MILLISECONDS.toHours(countDownVo.getMillisecondsFromDate());
+			} else if (countDownVo.getMillisecondsFromDate() >= Integer.MAX_VALUE) {
+				logger.error("Hours before take off was difference of *2,147,483,647+* hours! "
+						+ "This indicates a likely data issue!");
+				hoursBeforeTakeOff = Integer.MAX_VALUE;
```
simbamarufu1: Should this be ` TimeUnit.MILLISECONDS.toHours(Integer.MAX_VALUE)` ?

gtas-parent/gtas-loader/src/main/java/gov/gtas/services/GtasLoaderImpl.java
```diff
@@ -333,12 +335,46 @@ public PassengerInformationDTO makeNewPassengerObjects(Flight primeFlight, List<
...
+			transmissionDate = thisMessage.getEdifactMessage().getTransmissionDate();
+		}
+
+		if (transmissionDate != null && primeFlight.getMutableFlightDetails().getEtd() != null) {
+			Date flightDate = primeFlight.getMutableFlightDetails().getEtd();
+			CountDownCalculator countDownCalculator = new CountDownCalculator(transmissionDate);
+			CountDownVo countDownVo = countDownCalculator.getCountDownFromDate(flightDate);
+			if (countDownVo.getMillisecondsFromDate() > 0
+					&& countDownVo.getMillisecondsFromDate() <= Integer.MAX_VALUE) {
+				hoursBeforeTakeOff = (int) TimeUnit.MILLISECONDS.toHours(countDownVo.getMillisecondsFromDate());
```
simbamarufu1: Do we care that this truncates values vs rounding?

gtas-parent/gtas-loader/src/main/java/gov/gtas/services/GtasLoaderImpl.java
originalname51: It could either be negative or positive so I stuck with 0. This cases _should_ never happen so it is hopefully a non-issue.

gtas-parent/gtas-loader/src/main/java/gov/gtas/services/GtasLoaderImpl.java
originalname51: I like truncating more than rounding in this case. If a flight is .99 hours away then I would prefer 0 instead of 1.

tjwdev: It also makes it more understandable from a user perspective since .99 hours away is harder to grok.

PR #1755, opened 2020-03-18
Fixes for qbrb#1718

tjwdev:
This fixes current anomalous behaviour with In/Not in and directly related Equals/Not Equals with our query builder as it pertains to Email, Document, and Creditcard

The underlying issue is a combination of connecting certain elements (Email, Creditcard) through PNR rather than passenger, and that in any X to MANY relationship with our current query builder implementation the base line where clause construction is insufficient for not in and not equals specifically.

Going forward, IN and EQUALS will both function as soft inclusive, where NOT IN and NOT EQUALS will function as hard exclusive with their terms. This is not logically equivalent, but given our data structure as it is and how a user would feel about using it, this has been surmised as the correct way.

There is a lingering issue that exceeded the scope of this ticket in that selectize multi-variable selections seemed to concatenate values with a comma. This was fixed for the entities mentioned above, but an exhaustive search was not done for all potential string multi-values.

### DIFF
--- a/gtas-parent/gtas-commons/src/main/java/gov/gtas/querybuilder/JPQLGenerator.java
+++ b/gtas-parent/gtas-commons/src/main/java/gov/gtas/querybuilder/JPQLGenerator.java
@@ -319,6 +319,51 @@ public class JPQLGenerator {
 
 				}
 
+			} else if (entityEnum.toString().equalsIgnoreCase(Constants.EMAIL) ||
+					entityEnum.toString().equalsIgnoreCase(Constants.CREDITCARD) ||
+					entityEnum.toString().equalsIgnoreCase(Constants.DOCUMENT)) {
+
+				/* Due to a lack of direct connection between passenger and email/creditcard/etc...
+				 * IN/NOT_IN (and equals/not equals) behaviour does not work as intended through HQL, particularly because the order in
+				 * which the HQL separates the groupings goes from smallest -> largest, which allows for bad data
+				 * Example: A PNR 1, has 2 emails with domains: Gmail (2), and Hotmal (3). If Not In Gmail is used
+				 * 1 -> 2 would be false, but 1 -> 3 is true. This returns pnr 1 when it DOES contain gmail (which
+				 * in turn returns a passenger erroneously)
+				 * */
+
+				if (OperatorEnum.NOT_IN.toString().equalsIgnoreCase(operator)
+						|| OperatorEnum.NOT_EQUAL.toString().equalsIgnoreCase(operator)
+							) {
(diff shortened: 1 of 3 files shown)


gtas-parent/gtas-commons/src/main/java/gov/gtas/querybuilder/JPQLGenerator.java
```diff
@@ -319,6 +319,51 @@ private static void generateWhereCondition(QueryEntity queryEntity, EntityEnum q
 
 				}
 
+			} else if (entityEnum.toString().equalsIgnoreCase(Constants.EMAIL) ||
```
originalname51: Just to be clear - the entity enum is generated when the request comes in?

gtas-parent/gtas-commons/src/main/java/gov/gtas/querybuilder/JPQLGenerator.java
```diff
@@ -319,6 +319,51 @@ private static void generateWhereCondition(QueryEntity queryEntity, EntityEnum q
...
+
+				/* Due to a lack of direct connection between passenger and email/creditcard/etc...
+				 * IN/NOT_IN (and equals/not equals) behaviour does not work as intended through HQL, particularly because the order in
+				 * which the HQL separates the groupings goes from smallest -> largest, which allows for bad data
+				 * Example: A PNR 1, has 2 emails with domains: Gmail (2), and Hotmal (3). If Not In Gmail is used
+				 * 1 -> 2 would be false, but 1 -> 3 is true. This returns pnr 1 when it DOES contain gmail (which
+				 * in turn returns a passenger erroneously)
+				 * */
+
+				if (OperatorEnum.NOT_IN.toString().equalsIgnoreCase(operator)
```
originalname51: Can the operator ever be null?

tjwdev: The operator being null would be an incomplete query, right now currently query builder doesn't allow you to submit an incomplete query like that. It is technically possible to hit the end point behind it manually if you're logged in... But I 'm not sure that's reasonable.

gtas-parent/gtas-commons/src/main/java/gov/gtas/querybuilder/JPQLGenerator.java
tjwdev: Yes, entityEnum is generated as part of the query term for the query as a whole. The entityEnum MUST exist simply because it cannot be submitted through QB without it. (aside from manually hitting out endpoints)

PR #1781, opened 2020-04-24
Account locking after excessive login attempts

simbamarufu1:


gtas-parent/gtas-commons/src/main/java/gov/gtas/email/ResetPasswordEmailService 2.java
```diff
@@ -0,0 +1,42 @@
...
+    @Autowired
+    private GtasEmailService emailService;
+
+    @Transactional
+    public void sendAccountLockedResetEmail(String email, String resetToken) throws IOException, TemplateException, MessagingException, URISyntaxException {
+        String resetLinkUrl = new URIBuilder(resetLinkUrlBase).addParameter("token", resetToken).build().toString();
+        String htmlContent = emailTemplateLoader.generateAccountLockedResetPasswordHtmlContent(email, resetLinkUrl);
+
+        EmailDTO emailDTO = new EmailDTO();
+        emailDTO.setSubject(RESET_EMAIL_SUBJECT);
```
originalname51, line 36: making configurable might be overkill but i think it would add some value

gtas-parent/gtas-commons/src/main/java/gov/gtas/services/security/UserServiceImpl.java
```diff
@@ -159,9 +168,52 @@ public User fetchUser(final String userId) {
 		return userGroupRepository.findDistinctByGroupMembersIn(Collections.singleton(user));
 	}
 
+	@Override
```
originalname51, line 171: Did you consider breaking out a login service? It might be overkill but it would separate concerns.

gtas-parent/gtas-webapp/src/main/java/gov/gtas/controller/ResetController.java
```diff
@@ -0,0 +1,65 @@
...
+
+        if(isBlank(username)) {
+            return new JsonServiceResponse(Status.FAILURE, INVALID_CREDENTIALS_REQUEST_ERROR);
+        }
+
+        User user = userRepository.findOne(username);
+       if(user == null
+               || isBlank(resetToken)
+               || !StringUtils.equals(user.getResetToken(), resetToken)
+               || StringUtils.equals(password, user.getPassword())) {
```
originalname51, line 44: It might be worth it to update the html to say you cant use your old password.

gtas-parent/gtas-webapp/src/main/webapp/reset.html
```diff
@@ -0,0 +1,140 @@
+<!doctype html>
+<html lang="en" ng-app="myApp">
+
+<head>
+    <meta http-equiv="Content-Type" content="text/html;charset=utf-8">
+    <meta charset="utf-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <title>GTAS</title>
+    <link rel="stylesheet" href="resources/node_modules/npm-font-open-sans/open-sans.css" />
```
gtasautumn, line 10: Just fyi, this file, main, and login each duplicate references to most or all of the deps and style links for the entire app. It's a bad pattern. We could drop most of these at the top and bottom of this file. Could do it as a very tedious separate ticket.

gtas-parent/gtas-webapp/src/main/java/gov/gtas/security/SecurityUserDetailsService.java
```diff
@@ -44,10 +44,13 @@ public UserDetails loadUserByUsername(String username) throws UsernameNotFoundEx
 			throw new UsernameNotFoundException(message);
 		}
 
+		Integer loginAttempts = userService.getUserLoginAttempts(user.getUserId());
+		boolean accountNotLocked = loginAttempts == null || loginAttempts <= 5;
```
gtasautumn, line 48: For later, can we make this configurable rather than hard-coded?

PR #1789, opened 2020-05-08
Manual pvl generation

tjwdev:
Pull request - Still need to write unit tests for ticket completion, but can potentially be merged.


gtas-parent/gtas-commons/src/main/java/gov/gtas/services/PendingHitDetailsServiceImpl.java
```diff
@@ -0,0 +1,97 @@
...
+    @Autowired
+    HitCategoryService hitCategoryService;
+
+    //For generation of manual hits.
+    @Override
+    @Transactional
+    public void createManualPendingHitDetail(Long paxId, Long flightId, String userId, Long hitCategoryId, String desc){
+        User user = userService.fetchUser(userId);
+        String title = "Manually Generated PVL";
+        String ruleConditions = "N/A";
```
originalname51, line 41: Rule condition should be either input by the user or potentially given a default "Manually generated hit by user xxx"

gtas-parent/gtas-commons/src/main/java/gov/gtas/services/PendingHitDetailsServiceImpl.java
```diff
@@ -0,0 +1,97 @@
...
+        Float percentageMatch = 1F;
+
+        //This list should only return 1 given the current construction of our Manual generated HitMaker
+        List<ManualHit> mhList = hmr.findOneByHitCategoryIdAndHitType(hitCategoryId);
```
originalname51, line 45: This locks us in to category specific rules - probably ok or even good for now but enhancing this later on could be difficult (for instance 2 types of manual hits on a general category).

gtas-parent/gtas-commons/src/main/java/gov/gtas/services/PendingHitDetailsServiceImpl.java
```diff
@@ -0,0 +1,97 @@
...
+
+        return phd;
+    }
+
+    @Override
+    public void saveAllPendingHitDetails(Set<PendingHitDetails> phdSet){
+        phr.saveAll(phdSet);
+    };
+
+    private ManualHit createManualHitMaker(String desc, User user, Long hitCategoryId){
```
originalname51, line 84: is this dead code?

tjwdev: Arguably yes. I had originally put it in early because of potential use, but then didn't end up using it. Is removable.

gtas-parent/gtas-commons/src/main/java/gov/gtas/services/PendingHitDetailsServiceImpl.java
tjwdev: All hits require a category at the moment though, but you are correct, creating two general category manual hits will not generate the second, yes. But given we have no other current way to create a manual hit I'm thinking this should suffice  for now.

originalname51: 👍

gtas-parent/gtas-commons/src/main/java/gov/gtas/services/PendingHitDetailsServiceImpl.java
tjwdev: Will change to use userId and indicate was manually made. Can expand to have more detailed note later on the category selection sidebar.

PR #1799, opened 2020-05-27
Make docker-compose files suitable for kompose k8s tool

simbamarufu1:


TeddyTesfaye:
Looks good to me. As planned, the compose files will re-point to the team's repo once everything is tested.

PR #1858, opened 2020-07-21
#1857 Update seat to show class

originalname51:


gtasautumn:
Is there a way we can tag/flag PRs that have DB updates? that would be super helpful.

PR #1862, opened 2020-07-23
Addition of new delete capability

tjwdev:
-Added two new end points. One for collecting all non-archived users. One for attempting to delete users
-Added new user attribute "archived" boolean, database will need to be rebuilt.
-Modified relationship between user and user_group to enable removal from the corresponding join table
-Upon calling delete will attempt to delete the user, if fails for any reason it will attempt to archive the User.java

This must be merged for the sister pull request to work in gtas-ui

### DIFF
--- a/gtas-parent/gtas-webapp/src/main/java/gov/gtas/controller/UserController.java
+++ b/gtas-parent/gtas-webapp/src/main/java/gov/gtas/controller/UserController.java
@@ -13,6 +13,7 @@ import java.util.stream.Collectors;
 
 import javax.validation.Valid;
 
+import gov.gtas.services.security.RoleData;
 import org.passay.CharacterRule;
 import org.passay.EnglishCharacterData;
 import org.passay.LengthRule;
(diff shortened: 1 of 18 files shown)


gtas-parent/gtas-webapp/src/main/java/gov/gtas/controller/UserController.java
```diff
@@ -164,6 +169,29 @@ public JsonServiceResponse changePassword(@RequestBody PasswordChangeDTO passCha
 		
 		
 	}
+	@ResponseBody
+	@RequestMapping(method = RequestMethod.DELETE, value = "/users/{id}")
+	public JsonServiceResponse deleteOrArchiveUser(@PathVariable(value = "id") String paxId){
```
kasimhelil, line 174: I would suggest to have separate endpoints for **deleting** and **archiving** users. 
- We might need to archive a user with no constraints 
- If a user can't be deleted, tell the the _end user_  that they can't delete the user, instead  to attempt to archive it.

tjwdev: I'm not strictly opposed to having two end points for this, but there's two things I'm considering:

1. Is there a point that we want to a delete a user and yet don't want to have them archived otherwise?

2. Cluttering the user modal with multiple buttons, particularly two buttons that from their functional point of view do the same thing feels somewhat redundant.

kasimhelil: Good points! 
My point was having a separate end point would help us to archive a user (with no intention of deleting the user).

tjwdev: I do understand, and there is flexibility in that. Though as a fun thought excercise, is there a point that you'd want to archive someone who WOULDN'T be archived by trying to delete them? (I.E. a fresh user with no rules/notes/etc?) 

That said, this is more future-proofing in a way and certainly can be done. But if we are to future proof this way, would it then be wise to ALSO build an un-archive endpoint?

kasimhelil: I don't see any purpose to archive a fresh user for now, unless we want to reserve their data, e.g username.

PR #1864, opened 2020-07-27
User registration

kasimhelil:


tjwdev:
There was an interesting behaviour regarding making sign up requests, despite receiving a "Failed sending email message" both ended up being sent. See attached image: 
FailedSendUserRegistration

That said, the rest of the sign up process of both approving and rejecting one of each of the users functioned expectedly and I was able to successfully create users.

kasimhelil:
The error messages should be more descriptive, which I will fix right now. Since we are sending emails to the user and whoever is approving the request (for now is admin), it should tell where it failed. 
 But can you check if both the user and admin received email.

PR #1869, opened 2020-07-31
Manual lookout hotfix

simbamarufu1:


tjwdev:
Pulled and tested creating one of each type of manual hit from a clean database, no errors.

PR #1887, opened 2020-09-16
Admin/manageuser/updates

kasimhelil:
- Added phone number field for both user and signupRequest
- Prevent saving Admin role with other roles


gtasautumn (approved):
looks good

PR #1892, opened 2020-09-25
K8s/docker config changes

simbamarufu1:


remote-deployment.yml
```diff
@@ -51,11 +46,11 @@ services:
...
       - type: bind
-        source: /mnt/gtas-java-jre
-        target: /usr/local/gtas-java-jre
-      - type: bind
-        source: /var/opt/gtas_scheduler_application_properties
-        target: /usr/local/tomcat/conf
+      #   source: /mnt/gtas-java-jre
+      #   target: /usr/local/gtas-java-jre
+      # - type: bind
+      #   source: /var/opt/gtas_scheduler_application_properties
```
originalname51, line 52: Is it OK not to have gtas-scheduler app props in use?

simbamarufu1: this should be fine and can be changed during runtime

PR #1897, opened 2020-10-05
Resolves #221 on GTAS-UI side.

### LINKED ISSUE #221: Where does credit card holder come from in the PNR?

tjwdev:
-Update and rollback scripts for addition of view flights role

### DIFF
--- a/gtas-parent/scripts/db/gtas-ui_table_updates/re_add_view_flights_roles_update
+++ b/gtas-parent/scripts/db/gtas-ui_table_updates/re_add_view_flights_roles_update
@@ -1,18 +1,20 @@
 -- Add view flights role
-INSERT IGNORE into Role(role_id, role_description) VALUES(9,'View Flights');
+INSERT IGNORE into Role(role_description) VALUES('View Flights');
 -- Begin create procedure
 DROP PROCEDURE IF EXISTS ADDVIEWFLIGHTSROLE;
 DELIMITER ;;
---For every user that exists, give them the new user_role of id = 9
+--For every user that exists, give them the new view flights id
 Create Procedure ADDVIEWFLIGHTSROLE()
 BEGIN
+DECLARE roleId INT DEFAULT 0;
 DECLARE n INT DEFAULT 0;
 DECLARE i INT DEFAULT 0;
-DECLARE test VARCHAR(500);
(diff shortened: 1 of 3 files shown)


gtas-parent/scripts/db/gtas-ui_table_updates/re_add_view_flights_roles_update
```diff
@@ -0,0 +1,24 @@
+-- Add view flights role
```
gtasautumn, line 1: so the procedure is probably overkill since we are dropping it immediately, but no biggie.

First thing we should correct is the hard-coded role_id. In the event some installation has edited their DB, assuming the ids are unchanged is a risk, but it's easy to avoid. Insert View Flights and pull back the id as a variable, then you can safely use that for inserts.

``INSERT INTO ROLE (role_description) VALUES ("View Flights");``

``SELECT ROLE_ID INTO @vfrole
FROM ROLE WHERE ROLE_DESCRIPTION = 'View Flights';``

Second thing is to insert only for records that don't already have the role so it doesn't error if the script gets executed multiple times. Idempotence is extremely important when we don't know the current state of the client's DB:

``INSERT INTO USER_ROLE (SELECT USER_ID, @vfrole
FROM USER
WHERE USER_ID NOT IN (
	SELECT USER_ID
	FROM USER_ROLE
	WHERE ROLE_ID = @vfrole
));``

For the rollback, we should again pull the role_id into a variable and delete values in user_role using that variable.

This shd run in Heidi, but I have no idea how the orm will mangle it.

tjwdev: The reason I used a procedure is because making it run as a block execution was failing despite following MariaDB specific instructions, the same issue I'd had all those months ago with the pending_hits. 

I'll make the role_id programmatically generated for both, you're right that's a far better idea for safety's sake. Idempotency could potentially be solved with a unique key constraint? But checking in the script will be just as easy. I'll make those changes.

gtasautumn: Hadn't thought of that either. Could update it to exclude admins, but as long as it doesn't break anything we're OK.

gtasautumn (approved):

Issue #13, opened 2016-07-28
Flights View | Expand/Collapse funtion


dje04001:
It would be extremely useful to be able to expand on the flight number to expose the manifest, which could be sorted by any of the passenger fields, but default on hit / hit severity

dje04001:
Making the expansion limited to just rule/watchlist hits.

dje04001:
90 percent done on 9/16

PR #1906, opened 2020-10-14
Flight pax grid parity

tjwdev:
This is for #13 on GTAS-UI issues, as well as general flight-grid parity with hit display with the original UI.

This adds new co-traveller id to our VO's which consists of the pnr record locator number so that the front-end can group them.

This is the back-end pull request.

### DIFF
--- a/gtas-parent/gtas-commons/src/main/java/gov/gtas/services/PassengerServiceImpl.java
+++ b/gtas-parent/gtas-commons/src/main/java/gov/gtas/services/PassengerServiceImpl.java
@@ -119,8 +119,16 @@ public class PassengerServiceImpl implements PassengerService {
 					break;
 				}
 			}
-			for ( Pnr pnr : passenger.getPnrs()){
-				vo.setCoTravelerId(pnr.getId());
+			Pnr latestPnr = null; //grab most recent pnr (assumed to be most up to date)
+			Date mostRecentDate = null;
+			for(Pnr pnr : passenger.getPnrs()){
+				if(mostRecentDate == null) {
+					mostRecentDate = pnr.getDateReceived();
+					latestPnr = pnr;
+				} else if(mostRecentDate.before(pnr.getDateReceived())) {
+					latestPnr = pnr;
+				}//Flaw? Date booked better?
+				vo.setCoTravellerId(latestPnr.getRecordLocator());
 			}
 
 			// grab flight info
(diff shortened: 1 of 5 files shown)


gtas-parent/gtas-commons/src/main/java/gov/gtas/services/PassengerServiceImpl.java
```diff
@@ -305,6 +316,18 @@ public void setAllFlights(Set<Flight> flights, Long id) {
 				vo.addDocument(docVo);
 			}
 
+			Pnr latestPnr = null; //grab most recent pnr (assumed to be most up to date)
```
kasimhelil, line 319: Code duplication; I would rather have a `getMostRecentPnr(List pnrs)` method and use it here and on line 122

tjwdev: The reason there is code duplication is because one of the two vo's is slated to be axed as they serve the same purpose. This hasn't been decided yet, but it will eventually be pared down to one, removing the duplication as well.

kasimhelil: Got you! I should have noticed that!

gtas-parent/gtas-commons/src/main/java/gov/gtas/services/PassengerServiceImpl.java
```diff
@@ -119,6 +119,17 @@ public PassengersPageDto getPassengersByCriteria(Long flightId, PassengersReques
...
 				}
 			}
+			Pnr latestPnr = null; //grab most recent pnr (assumed to be most up to date)
+			Date mostRecentDate = null;
+			for(Pnr pnr : passenger.getPnrs()){
+				if(mostRecentDate == null) {
+					mostRecentDate = pnr.getDateReceived();
+					latestPnr = pnr;
+				} else if(mostRecentDate.before(pnr.getDateReceived())) {
+					latestPnr = pnr;
```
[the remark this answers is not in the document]
tjwdev: You're absolutely correct, applying fix now.

kasimhelil (approved):

gtasautumn (approved):

gtasautumn:
Ping us when you have the latest fix in

Issue #248, opened 2017-03-13
DDU - "in transit passenger" | APIS data


dje04001:
Research whether or not this is parsed/loaded, if it's worth parsing and loading it.

svempati2006:
We need to receive the close out APIS message to find the previously loaded passengers based on the reference numbers/record numbers and update their status to "Boarded" or "in Transit"
Pl see the API documentation for details.


5.5 Flight Close-Out PAXLST Message
(**B.Blankenship – inclusion of this FCO example in the Annex is at the discretion of the API CC
Committee.)**
This sample illustrates a PAXLST message that may be used to report a Flight Close-Out
message that identifies all passengers who boarded the flight identified in the message.
In this example, passengers are identified only by Passenger Recor Locator and Unique
Passenger Reference (RFF qualifiers 'AVF' and 'ABO' respectively). The expectation, in this
example, is that the passenger names (travel documentation, etc.) were collected through
previously transmitted PAXLST API submissions.
UNB+UNOA:4+XYZ+CUSTOMS+130322:0335+0000001++API'
UNG+PAXLST+XYZ AIRLINES+CUSTOMS+130322:0335+1+UN+D:12B'
UNH+5755176+PAXLST:D:12B:UN:IATA'
BGM+266+CLOB'
RFF+TN:ABC1234:::1'
TDT+20+YZ567+++AA'
LOC+125+LHR'
DTM+189:1303221615:201'
LOC+87+LAX'
DTM+232:1303221905:201'
NAD+ZZZ'
RFF+AVF:TYR123’
RFF+ABO:TYL001’
NAD+ZZZ'
RFF+AVF:TYR123’
RFF+ABO:TYL002’
NAD+ZZZ'
RFF+AVF:TYR123’
RFF+ABO:TYL003’
NAD+ZZZ'
RFF+AVF:TYR123’
RFF+ABO:TYL004’
NAD+ZZZ'
RFF+AVF:AABD55’
RFF+ABO:MCO001’
NAD+ZZZ'
RFF+AVF:AABD55’
RFF+ABO:MCO002’
NAD+ZZZ'
RFF+AVF:ZMJO6O’
RFF+ABO:VEF001’
CNT+42:7’
UNT+31+5755176'
UNE+1+1'
UNZ+1+0000001’

Also see the NAD segment which has a Value :ZZZ which will not give us a clue whether passenger is "In Transit" or not.
The NAD in group 4 will intimate if the passenger is in transit..
[...]

Issue #139, opened 2016-10-30
Passenger Details | add to Watch List button | 5 points


dje04001:
add a new button, "Add to Watch List" and use the glyphicon "eye-plus"

dje04001:
services exist to call back end.

dje04001:
done, but not committed

PR #1910, opened 2020-10-19
Edit delete hit note cats

### LINKED ISSUE #325: Parsing and Loading | Citizenship
The citizenship field isn't being populated in the passengers view by PNR when it arrives before APIS message

tjwdev:
Resolves #325, #248, #139 ON GTAS-UI SIDE.

This has  sister front-end PR with the same branch name. Database update is required.


tjwdev:
After much deliberation and testing, new Hit Categories will effectively always be archived as the current relationships are built. Functionally this keeps them on record, form wise however it will be as if they were deleted on the front-end. 

Revisiting the relationships between the trouble areas restricting deletion should be done-so in the future.

gtasautumn (approved):
looks good

PR #1913, opened 2020-10-20
Clean up application statistics data

kasimhelil:
- created a new dto that consists only fields required by the front end

- Updated all returned date types to Long so that the front end can
parse them correctly

- Updated total loading error to include DUPLICATE_MESSAGE (this was
throwing error on the prev code and the fetch was failing)


gtasautumn (approved):
looks good

PR #1917, opened 2020-10-29
Backend update to allow for sending of a passenger.

originalname51:
This will allow the gui to request that the rule engine send a passenger, note, and latest relevant message to another system utilizing the gtas-information-share pojos.

This also incorporates new parsing to parse in passengers who contain notes.


tjwdev (approved):
lgtm

PR #1969, opened 2020-11-23
Audit log fixes

### LINKED ISSUE #428: Query Builder | Baggage Changes
refer to ticket 427 regarding baggage changes

tjwdev:
Sister backend PR for AuditLogFixes branch on GTAS-UI
Fixes #428 on GTAS-UI side
-Pulled actions from enumerated list
-Fixed dates not working
-Fixed ALL_ACTIONS showing up and working correctly
-Removed base "Select Action" from drop down
-Merged Kasim's changes for re-adding OrderBy for search


tjwdev:
This now also resolves #429 on GTAS-UI side.

gtas-parent/gtas-webapp/src/main/java/gov/gtas/controller/AdminController.java
```diff
@@ -363,7 +362,9 @@ public int restoreAllCreditCardType() {
 	    
 	    if (actionType != null ) {
 	    	params.put("actionType", actionType);
-	    }
+	    } else {
+	      params.remove("actionType"); //Correctly allows for search criteria to be built for ALL_ACTIONS.
```
gtasautumn, line 366: So this works, BUT if we are just removing it here wouldn't it be easier just not to send the code from the frontend? I noticed it because including it as an enum causes the `ALL_ACTIONS` option to be ordered beneath `ACCESS_INFORMATION` . If the backend really doesn't need the ALL tag, its simpler for the front-end to append a dummy option at the top of the dropdown list to represent the ALL option and just not send it when the default is selected.

gtasautumn (changes requested):
Everything works, but the sort here prevents ALL_ACTIONS from displaying at the top of the dropdown on the frontend.

gtas-parent/gtas-webapp/src/main/java/gov/gtas/controller/AdminController.java
kasimhelil: Making action's input not required on the front end should be enough, I think. 
And then we can remove ALL_ACTIONS enum from both the frontend and backend. 
Not selecting any action types means give me all; right?

gtas-parent/gtas-webapp/src/main/java/gov/gtas/controller/AdminController.java
```diff
@@ -372,11 +373,16 @@ public int restoreAllCreditCardType() {
...
-	   
-	    
+
 	    return auditService.getAuditlog(params);
   }
+
+  @RequestMapping(method = RequestMethod.GET, value = "/api/auditlog/actions")
+  public List<AuditActionType> getActionTypes(){
+    List<AuditActionType> actionList = Arrays.asList(AuditActionType.values());
+    actionList.sort(Comparator.comparing(AuditActionType::name));
```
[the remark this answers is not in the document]
tjwdev: The solution will be applied as such: 
1. Front-end will generate the "ALL_ACTIONS"/"ALL_CODES" and put at the top of the list. (Force "ALL" to always be top option)

2. Front-end will not send the parameter for action/code if those are the values. (Remove clunky-parameter removal)

gtas-parent/gtas-webapp/src/main/java/gov/gtas/controller/AdminController.java
tjwdev: I think all actions (and codes) is a valuable thing for the user to see/use, so I'm not so certain on the front-end it needs to be removed. But having it never get sent back as an actual parameter will work in order to remove it from the backend enum, which ,agreed, it does not need to be in.

gtas-parent/gtas-webapp/src/main/java/gov/gtas/controller/AdminController.java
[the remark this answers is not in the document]
kasimhelil: I am for removing the `ALL_ACTIONS`/`ALL_CODES` options and setting required to false. 
And just have a place holder, something like `Select actions` /`Select codes`

gtasautumn (approved):

PR #1976, opened 2020-12-10
Fix/user roles

gtasautumn:
- Removed Manage Cases role
- Roles entity immutable
- User Roles validated against DB - currently matched on description only but more updates are necessary to remove hard-coded references.
- misc refactoring of redundant code.

So the updates affect the create and update paths on the User entity. Need to verify that admins/non-admins can change their own password successfully under the site header, and that admins can:
- edit the roles of other users such that View Flights is always saved with the collection of selected non-admin roles, or that the Admin role is saved by itself no matter what other roles were selected.
- Edit the password of other users as expected
- Create a user as expected.

The issue with creating invalid roles can only be tested by changing the default role id in UserModal.js to 1 and the description to any random text. Then if you attempt to save any combination of roles excluding admin, you should see that the invalid one does not get created in the roles table or associated with the user when you go back to edit them.


tjwdev (approved):
LGTM

PR #1988, opened 2021-01-06
Save And Run Rules on Segments

originalname51:
Some of the information GTAS ingest from PNR data consist of free text fields.

It appears that these fields could follow some degree of structure, but this is not defined or enforced in any meaningful way. Because of this, there is no way to reliably extract contact information or other informative pieces.

To allow for more of a manual search of this certain segments, the parser has been enhanced to take a copy of none to all of the segments IFT,SAC,LTS,SSR,EBD, and ADD. To allow selectivity when selecting segments they must match a list of provided regex. Both the segments and the provided regex are defined in the default.application.properties. Note that only the first matching regex will be recorded by the saved segment. By default no segments are saved.

Once the segments have been extracted they are saved to the saved_segments table of GTAS. Rules and queries can be run on the raw messages, the segment name, and the regex that caused the segment to be saved.

Due to the nature of the free text the only query and rule functionality provided is "contains" and "not contains."

### DIFF
--- a/gtas-parent/gtas-loader/src/main/java/gov/gtas/services/GtasLoaderImpl.java
+++ b/gtas-parent/gtas-loader/src/main/java/gov/gtas/services/GtasLoaderImpl.java
@@ -200,28 +200,27 @@ public class GtasLoaderImpl implements GtasLoader {
 		pnr.setPhones(null);
 		pnr.setDwellTimes(null);
 		pnr.setPaymentForms(null);
-		String stacktrace = ErrorUtils.getStacktrace(e);
-		pnr.setError(stacktrace);
+		pnr.setError(e.toString());
 		if (e instanceof DuplicateHashCodeException) {
 			logger.info(e.getMessage());
 			pnr.getStatus().setMessageStatusEnum(MessageStatusEnum.DUPLICATE_MESSAGE);
 		} else {
 			pnr.getStatus().setMessageStatusEnum(MessageStatusEnum.FAILED_LOADING);
-			logger.error(stacktrace);
+			logger.error("Error in gtas Loader", e);
 		}
 	}
 
 
 	public static void handleException(Exception e, ApisMessage apisMessage) {
 		String stacktrace = ErrorUtils.getStacktrace(e);
-		apisMessage.setError(stacktrace);
+		apisMessage.setError(e.toString());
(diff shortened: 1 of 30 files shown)


gtas-parent/gtas-loader/src/main/java/gov/gtas/services/GtasLoaderImpl.java
```diff
@@ -200,28 +200,27 @@ public static void handleException(Exception e, Pnr pnr) {
...
-		String stacktrace = ErrorUtils.getStacktrace(e);
-		pnr.setError(stacktrace);
+		pnr.setError(e.toString());
 		if (e instanceof DuplicateHashCodeException) {
 			logger.info(e.getMessage());
 			pnr.getStatus().setMessageStatusEnum(MessageStatusEnum.DUPLICATE_MESSAGE);
 		} else {
 			pnr.getStatus().setMessageStatusEnum(MessageStatusEnum.FAILED_LOADING);
-			logger.error(stacktrace);
+			logger.error("Error in gtas Loader", e);
```
tjwdev, line 209: Inconsistent error message GTASLOADER vs gtas Loader

gtas-parent/gtas-rulesvc/src/main/java/gov/gtas/rule/builder/pnr/PnrRuleConditionBuilder.java
```diff
@@ -90,6 +91,12 @@ public PnrRuleConditionBuilder(List<QueryTerm> queryTermList) {
 				conditionBuilderMap.put(uuid, ecb);
 				phoneConditionBuilder.add((PhoneConditionBuilder) ecb);
 				break;
+			case SAVED_SEGMENT:
+				groupNumber = conditionBuilderMap.size() + 1;
+				ecb = getEntityConditionBuilder(entityEnum, groupNumber);
+				conditionBuilderMap.put(uuid, ecb);
+				segmentConditionBuilder.add((SegmentConditionBuilder) ecb);	ecb.setGroupNumber(groupNumber);
```
tjwdev, line 98: Odd line breaking

tjwdev (approved):
Looks good to me, barring few clerical errors that don't strictly have to be fixed.

PR #1990, opened 2021-01-08
Updating email reset properties for the new UI

simbamarufu1:
1. Update reset properties for new UI which uses different endpoints than the legacy UI.
2. Rename property names to reflect that the proxy is now part of the UI.
3. Codify email related properties


kasimhelil (approved):
Looks good to me!!

PR #2000, opened 2021-01-14
Feat/context

gtasautumn:
backend updates for the context fetches.
Ran organize imports on the webapp java source
Removed the 8080 webapp


kasimhelil (approved):
Tested it locally and looks good to me!!!

PR #2007, opened 2021-01-22
Forgot username

kasimhelil:
- You need to set up the email service properties  on application properties file.
- Have a valid  email saved on db for the user you want test this functionality


gtas-parent/gtas-commons/src/main/java/gov/gtas/services/security/UserServiceImpl.java
```diff
@@ -287,6 +299,28 @@ public void forgotPassword(User user) {
 		
 	}
 	
+	@Override
+	@Transactional
+	public void forgotUsername(String userEmail) throws NotFoundException, MessagingException, IOException, TemplateException {
+		
+		String userId = userRepository.findUserByEmail(userEmail);
+		
+		if (userId == null) {
```
gtasautumn, line 308: Everything works, but the code constructing the email should be handled in the same class as for forgot password since we are already using that pattern. We should also rename that class, ResetPasswordEmailService to something more generic like "UserEmailService" or just "EmailService". The current location of ResetPasswordEmailService outside the "services" directory is also questionable, but that's a separate issue.

gtas-parent/gtas-commons/src/main/java/gov/gtas/services/security/UserServiceImpl.java
```diff
@@ -234,8 +245,9 @@ public boolean treatAsOneDay(String userId) {
 		Update all user data except the password. The password is handled by a separate function
 	 */
 	@Override
+	@Transactional
 	public UserData updateByAdmin(UserData data) {
-		User entity = userRepository.findOne(data.getUserId());	//returns error if null
+		User entity = fetchUser(data.getUserId());	//returns error if null
```
gtasautumn, line 250: this code is old. This was reverted back to userRepo.findOne to fix the issue with deleted user_groups. Here and in updatePassword.

gtasautumn (changes requested):
3 changes required, see comments

gtas-parent/gtas-commons/src/main/java/gov/gtas/services/security/UserServiceImpl.java
```diff
@@ -287,6 +299,28 @@ public void forgotPassword(User user) {
...
 	}
 	
+	@Override
+	@Transactional
+	public void forgotUsername(String userEmail) throws NotFoundException, MessagingException, IOException, TemplateException {
+		
+		String userId = userRepository.findUserByEmail(userEmail);
+		
+		if (userId == null) {
+			String message = "The provided email (" + userEmail + ") is not on the system!";
```
gtasautumn, line 309: Everything works, but the code constructing the email should be handled in the same class as for forgot password since we are already using that pattern. We should also rename that class, ResetPasswordEmailService to something more generic like "UserEmailService" or just "EmailService". The current location of ResetPasswordEmailService outside the "services" directory is also questionable, but that's a separate issue.

PR #2009, opened 2021-01-27
Add default check for 4 numbers on an IFT

originalname51:


gtas-parent/gtas-commons/src/main/resources/default.application.properties
```diff
@@ -280,5 +280,5 @@ omni.derogReplace.batchSize=300
 # Title and description to be used on derog hits received from Omni.
 omni.derog_hit.title=Kaizen External Hit
 omni.derog_hit.description=This passenger closely matches at least one derog category, according to Kaizen's proprietary inference technology.
-segments.extraparse=
-segments.regex=
+segments.extraparse=IFT
+segments.regex=.*[0-9]{4}.*
```
gtasautumn, line 284: So this regex capture group will be the 4 digit number and all surrounding text? Is it identifying that pattern as "of interest"? What's the context here?

originalname51: This will save an IFT segment containing 4 numbers in a row. This is typical in the address field of the test data we generate and can be used to validate this ticket. 

Our test environment is set up to have our values from the application properties hard coded.

tjwdev (approved):
Looks good to me!

PR #2012, opened 2021-02-08
Fix/seat charts

kasimhelil:


gtas-parent/gtas-commons/src/main/java/gov/gtas/services/FlightServiceImpl.java
```diff
@@ -285,6 +285,7 @@ public int getPassengerCount(Flight f) {
 				vo.setPaxId(passenger.getId());
 				vo.setFirstName(passenger.getPassengerDetails().getFirstName());
 				vo.setLastName(passenger.getPassengerDetails().getLastName());
+				vo.setMiddleName(passenger.getPassengerDetails().getMiddleName());
 				vo.setMiddleInitial(passenger.getPassengerDetails().getMiddleName());
```
gtasautumn, line 289: looks good! There is now a middleName and a middleInitial field at this line, both assigned the same data. I wonder if that was the issue. We should remove the middleInitial field if its not being used, or if it is replace it with middleName.

gtasautumn (approved):
looks good.

PR #2013, opened 2021-02-17
Query Performance Updates

originalname51:
* Added several commonly used indexes to increase speed of queries.
Added the annotations to classes with indexes. These include the
following
    - Composite index on passenger details (dob - gender - last name -
first name
    - Phone number index
    - Composite index on credit card (type - number)
    - index on booking detail origin
    - index on booking detail destination
    - index on document number
    - index on dwell time arrival airport
    - Composite index on email (domain, address)
* Updated JPQL generator to use inner joins on "and" statements
* Updated age to use dob on back-end to take advantage of new index on
passenger detail.
* Updated JPQL generator to in-take a query that has been decorated with
meta-data which includes UTC offset in minutes from requester origin.
* Updated unit test


tjwdev (approved):
Performed code review, looks good.

PR #2049, opened 2021-05-07
Seat service fix to allow for a passenger to hold multiple seats on a…

### LINKED ISSUE #726: neo4j | keylines
https://neo4j.com/blog/keylines-graphing-github/

tjwdev:
… given flight

-Change to vo's and base service to enable multiple seats to be returned for a given passenger rather than originally null if > 1 seat
-QueryBuilder will select first seat for passengergridVO return (though it is currently not used on the front-end)

Fixes #726 on the GTAS UI side


gtasautumn (approved):
looks good

Issue #784, opened 2018-04-09
PNR Tab | raw PNR not showing latest


dje04001:

svempati2006:
Code completed unit tested and checked into Dev barnch.Pl review.

PR #2055, opened 2021-05-26
Fix/vetting hits results

gtasautumn:
PVL DTO now returns the full hit detail text. Removed unused fields, methods and supporting architecture to reduce the size of the resultset.

See frontend PR [#784](https://github.com/US-CBP/GTAS-UI/pull/784)


tjwdev (approved):
LGTM.

PR #2056, opened 2021-05-27
Adding new enum and allowing PWL hits to be seen on POE

tjwdev:
-PWL was ignored as POE lookout due to their copious volume, added them back to display
-Demoted has been added as an enum to differentiate between automatic removal from POE vs having never been promoted to it in the first place.

Also adds a time buffer to "Missed" flight to give leeway for "Missing" arrivals or departure.

There is a PR on the front end with branch "POEDemotedEnumAndPWLHitDisplayFix" is required to guarantee accurate vetting page results.

### DIFF
--- a/gtas-parent/gtas-commons/src/main/java/gov/gtas/repository/HitViewStatusRepository.java
+++ b/gtas-parent/gtas-commons/src/main/java/gov/gtas/repository/HitViewStatusRepository.java
@@ -43,8 +43,8 @@ public interface HitViewStatusRepository extends CrudRepository<HitViewStatus, L
             "left join fetch hd.hitMaker hm "
     + " where hvs.userGroup in :userGroups "
     + " AND NOT hvs.hitViewStatusEnum = 'REVIEWED' "
-    + " AND NOT hvs.hitDetail.hitType = 'PWL' "
     + " AND NOT hvs.lookoutStatusEnum = 'NOTPROMOTED' "
+    + " AND NOT hvs.lookoutStatusEnum = 'DEMOTED' "
     + " AND (mf.etd BETWEEN :startDate AND :endDate "
     + " OR mf.eta BETWEEN :startDate AND :endDate)")
     Set<HitViewStatus> findAllWithNotClosedAndWithinRange(@Param("userGroups") Set<UserGroup> userGroups,
--- a/gtas-parent/gtas-commons/src/main/java/gov/gtas/repository/HitViewStatusRepository.java
+++ b/gtas-parent/gtas-commons/src/main/java/gov/gtas/repository/HitViewStatusRepository.java
@@ -42,11 +42,10 @@ public interface HitViewStatusRepository extends CrudRepository<HitViewStatus, L
             "left join fetch p.hitDetails hd " +
             "left join fetch hd.hitMaker hm "
     + " where hvs.userGroup in :userGroups "
-    + " AND NOT hvs.hitViewStatusEnum = 'REVIEWED' "
     + " AND NOT hvs.lookoutStatusEnum = 'NOTPROMOTED' "
     + " AND NOT hvs.lookoutStatusEnum = 'DEMOTED' "
-    + " AND (mf.etd BETWEEN :startDate AND :endDate "
-    + " OR mf.eta BETWEEN :startDate AND :endDate)")
+    + " AND ( (f.direction = 'O' AND mf.etd BETWEEN :startDate AND :endDate) "
+    + " OR (f.direction ='I' AND mf.eta BETWEEN :startDate AND :endDate ))")
     Set<HitViewStatus> findAllWithNotClosedAndWithinRange(@Param("userGroups") Set<UserGroup> userGroups,
                                                            @Param("startDate") Date startDate, @Param("endDate") Date endDate);
 }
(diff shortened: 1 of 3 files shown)


tjwdev:
For this state of the POE Lookout there are a few work flows to explain.

**1. What are the ways a Passenger can be "Promoted To Lookout"?**
- Through automatic promotion via hit category attribute
- Through manual promotion via the vetting page's PVL management, or on the Passenger Detail Page via Manual Hit

**2. What is the automatic promotion?**
- Hit categories now have a new boolean attribute that can be toggled active. 
- Under the hit categories management under the admin tab editing or creating a new hit category will allow you to select "Enable Automatic Lookout Promotion"
- Any rule or watchlist using this category will set the Passenger it hits upon to "ACTIVE" status on the POE Lookout page.
Note: This will not retroactively make "ACTIVE" any passengers previously hit, only future hits from the given rule with the applied category.

**3. What is manual promotion?**
- You may manually promote a passenger that has been hit and is not already lookout status "ACTIVE" by going to the vetting page and from the "Choose Action" drop down on a given row, select "Promote To Lookout" to set the given passenger's lookout status "ACTIVE".
Note: If a lookout status is not "NOTPROMOTED" or "DEMOTED", the option to promote the passenger will not be available.
- You may also promote a person without a hit by applying a Manual Hit on the Passenger Detail page of the given passenger, and selecting a category for the Manual Hit that is toggled with "Automatic Lookout Promotion".

**4. There is an automatic inactivity flow on the POE lookout**

- There are 3 types of automatic movement on the POE. 
- If the Lookout is not "NOTPROMOTED", "DEMOTED", "INACTIVE", or "MISSED" after a set amount of time of not having an update on the hitViewStatus it will automatically be moved to "INACTIVE"
- If the Lookout is "INACTIVE" for a certain amount of time, it will be automatically be moved to "DEMOTED"
- If the Lookout's flight has either departed or arrived (inbound/outbound depending) after a certain buffer period, it will be automatically be moved to "MISSED".
- If the Lookout status is updated AFTER it has been set to "MISSED", it will not be reset to "MISSED" automatically.
  
  To summarize: 
  <Any Non Inactive/NotPromoted/Demoted/Missed Status> --after no update time--> Inactive
  <Any Non Inactive/NotPromoted/Demoted/Missed Status> --after flight missed buffer time--> Missed
  Inactive --after inactivity time--> Demoted

  These variables are located in the POEServiceImpl, they are called:
   inactiveWindowVariable, fullyDemoteVariable, missedFlightBufferVariable. 
  
  Adjusting their date times will affect how quickly or slowly programmatic progression of the status' go.

**5. You may edit the POE Lanes**
- Under the admin tab exists a new POE Lane management page
- You may edit the visibility, order, and name of the POE Lanes.

gtas-parent/gtas-commons/src/main/java/gov/gtas/services/POEServiceImpl.java
```diff
@@ -48,9 +48,6 @@
     @Autowired
     private UserService userService;
 
-    private static final DateTime inactiveWindowVariable = new DateTime().minusDays(1);
-    private static final DateTime fullyDemoteVariable = new DateTime().minusWeeks(1);
-
     @Override
     public Set<LookoutStatusDTO> getAllTiles(String userId, POETileServiceRequest request) {
```
gtasautumn, line 52: This function is fetching the tile data for display, but it's also potentially updating the statuses simultaneously:

``hitViewStatusRepository.saveAll(hvsToBeUpdated);``

From experience, structures like this are troublemakers. Is there a reason we are not handling state updates in a scheduled job, or even calculating them in a view or code rather than saving them to disk? The issue is there's no safeguard preventing other code from bypassing the update. Any ETL, new get handler, or raw query that pulls this data directly from the DB would get bad statuses since the ONLY way to ensure good data now is to trigger an update with a front-end request. It will cause issues down the line.

Other thing is that these updates are triggered by a GET which should never mutate the DB outside of an audit log or other meta table. Never the data it's fetching.

We will need come up with a more robust solution for this.

gtas-parent/gtas-commons/src/main/java/gov/gtas/repository/HitViewStatusRepository.java
```diff
@@ -42,11 +42,10 @@
...
             "left join fetch hd.hitMaker hm "
     + " where hvs.userGroup in :userGroups "
-    + " AND NOT hvs.hitViewStatusEnum = 'REVIEWED' "
     + " AND NOT hvs.lookoutStatusEnum = 'NOTPROMOTED' "
     + " AND NOT hvs.lookoutStatusEnum = 'DEMOTED' "
-    + " AND (mf.etd BETWEEN :startDate AND :endDate "
-    + " OR mf.eta BETWEEN :startDate AND :endDate)")
+    + " AND ( (f.direction = 'O' AND mf.etd BETWEEN :startDate AND :endDate) "
+    + " OR (f.direction ='I' AND mf.eta BETWEEN :startDate AND :endDate ))")
     Set<HitViewStatus> findAllWithNotClosedAndWithinRange(@Param("userGroups") Set<UserGroup> userGroups,
```
gtasautumn, line 51: transactional?

tjwdev: There's no existing session for when that gets called, @transactional is used to open a new session with the DB for that query. I didn't put it on the parent method because it was a larger scope than required.

gtasautumn (approved):

Issue #797, opened 2018-05-01
Dev Poor Performance | queries


dje04001:
tried to search for a specific passenger by last name, query never returned expected results. Likely caused by incomplete pax tag ID.

Issue #786, opened 2018-04-10
PNR Loader | association and save profiling


dje04001:

svempati2006:
Can you elaborate what needs to be done  under (association and save profiling)?

dje04001:
Sure thing. Profiling each individual step to identify what exactly is taking the most time, and which steps can be improved. Jon said that the association and saving steps are taking the most time, and that there may be a more efficient way to associate passengers with flights.

PR #2057, opened 2021-06-01
Fix/lookup data updates

gtasautumn:
Fixes frontend bug [#797](https://github.com/US-CBP/GTAS-UI/issues/797) and [#786](https://github.com/US-CBP/GTAS-UI/issues/786)

Updates country and airport codes to correct missing country data in the airport dataset. DB scripts only, no code. 

## To setup locally
Because the iata codes for airports aren't guaranteed unique, we reliably can't do insert/insert ignore with the code as the unique key if you have modified your data. We likewise can't rely on the id field as identifiers because those values change in the non-restore tables if you've done a restoreAll.

You can either rebuild your DB from scratch or 

For countries:
1. run the country script /scripts/db/2.0/lookupDataUpdates.sql
2. in code editor, click **restore all** for countries. 

If you run into issues, you can do a full table restore:
1. disable the FK in country_and_organization 
2. truncate country and country_restore 
3. run the country insert script /gtas-commons/src/main/resources/sql/country.sql

For Airports, just do the full restore:
1. Truncate your airport and airport_restore tables
2. Run the Airport data insert script /gtas-commons/src/main/resources/sql/airports.sql

Once you have updated the codes, you may see duplicates in the code editor that are not reflected in your database. This is likely because of the way we handle duplicates in your local indexeddb. To resolve it:
1. Open Dev tools in tab running gtas
2. Under Application > Storage > IndexedDB > LookupDb, click Delete Database
3. Close the browser tab and login again in a new tab

image

## To Test
Go to countries in code editor and 
1. verify there are now entries for iso3 codes CUW, BES, SXM
2. under Airports, sort the data by country asc and desc and verify that there are no nulls for country


kasimhelil (approved):
looks good to me

Issue #807, opened 2018-05-25
Scheduler Compression Job | Frequent Flyer Numbers, Credit Card


dje04001:

originalname51:
Compression job removed, cc and ffn are linked to flight now.

PR #2062, opened 2021-06-16
Added pos/neg feedback loop for POE

### LINKED ISSUE #806: jobScheduler.properties | max number of files
We should never set this to 5. This killed the demo today.

tjwdev:
-Adjusted positive/negative to be immune to inactivity/demoted check
-added new elements to udrlistelement to include new count

fixes #806 and #807 on GTAS-UI side.


gtasautumn (approved):
we should try to make the count fetches more efficient if possible, (we now have 3 and the 2 new ones make the exact same call twice that could be made once with a param collection) but there is an upcoming ticket to expand this functionality that could make that issue obe anyway. Merging it as is.

PR #2063, opened 2021-06-29
Added new fields to attachmentVo

tjwdev:
Fixes frontend [570](https://github.com/US-CBP/GTAS-UI/issues/570)

-passing new fileSize and createdAt dates for view on attachments
Has sister PR [837](https://github.com/US-CBP/GTAS-UI/pull/837) on GTAS front end

### DIFF
--- a/gtas-parent/gtas-commons/src/main/java/gov/gtas/vo/passenger/AttachmentVo.java
+++ b/gtas-parent/gtas-commons/src/main/java/gov/gtas/vo/passenger/AttachmentVo.java
@@ -10,7 +10,16 @@ public class AttachmentVo {
 	private String filename;
 	private byte[] content;
 	private String contentType;
-	private Date created;
+	private Date createdAt;
+	private float fileSize;
+
+	public float getFileSize() {
+		return fileSize;
+	}
+
+	public void setFileSize(float fileSize) {
+		this.fileSize = fileSize;
+	}
 
 	public Long getId() {
 		return id;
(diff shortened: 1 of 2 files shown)


gtas-parent/gtas-webapp/src/main/java/gov/gtas/controller/UploadController.java
```diff
@@ -183,6 +183,8 @@ private void storeAttachmentToPassenger(MultipartFile file, String desc, String
 			attVo.setContentType(a.getContentType());
 			attVo.setDescription(a.getDescription());
 			attVo.setFilename(a.getFilename());
+			attVo.setCreatedAt(a.getCreatedAt());
+			attVo.setFileSize((a.getContent().getBytes(1, (int) a.getContent().length()).length)/1024);
```
gtasautumn, line 187: We are converting to KB automatically here, which is different than the conversion displayed in the attachment modal on the frontend. Here if I upload a file under 1kb or over a meg, the modal shows the size dynamically:
image

But on the grid, its always in KB because the backend is forcing a single format on us, and it rounds sizes under 1 KB to 0, which is incorrect:
image

We already have a util (formatBytes) on the front end which is handling conversions nicely, so we should return the size in bytes here and handle the conversions frontend for consistency.

gtasautumn (changes requested):
see comments

gtas-parent/gtas-commons/src/main/java/gov/gtas/vo/passenger/AttachmentVo.java
```diff
@@ -10,7 +10,16 @@
 	private String filename;
 	private byte[] content;
 	private String contentType;
-	private Date created;
+	private Date createdAt;
```
gtasautumn, line 13: We also need `Created By` here

gtasautumn (approved):
looks good

PR #2064, opened 2021-06-29
Refactored part of flighthistory request

tjwdev:
-Removed erroneous fields from booking detail history
-Adjusted query to consider a date recall limit on how far the history reaches (set to 1 year tentatively), also considers flight direction.
-added datetime to history call to pass to new query


gtas-parent/gtas-commons/src/main/java/gov/gtas/repository/BookingDetailRepository.java
```diff
@@ -22,10 +22,15 @@
 	@Query("SELECT pax FROM Passenger pax " + "left join fetch pax.passengerDetails  "
 			+ "left join fetch pax.bookingDetails "
 			+ "left join fetch pax.passengerTripDetails "
-			+ "left join fetch pax.flight "
-			+ "WHERE pax.id IN ("
+			+ "left join fetch pax.flight f "
+			+ "left join fetch f.mutableFlightDetails mf "
+			+ "WHERE (f.direction = 'O' AND mf.etd > :dateLimit) "
+			+ "OR (f.direction ='I' AND mf.eta > :dateLimit ) "
+			+ "OR (f.direction = 'A' AND (mf.etd > :dateLimit "
```
originalname51, line 29: f.direction of 'C' is not in place here.

PR #2075, opened 2021-08-16
Add horizontal loading.

originalname51:
Update GTAS to allow for horizontal loading, instead of just vertical.
Break out loading logic into loader workers.
Keep scheduler as the mediator and first point of contact for all
messages.
It is critical that GTAS only process a flight on a single thread across
all loaders. The scheduler is responsible for ensuring all loader
workers have access to different flights, and none are being worked at
the same time.


docker-compose.yml
```diff
@@ -109,7 +109,38 @@ services:
     environment:
       KB_LIST: UDR Knowledge Base_1,WL_Knowledge_Base_3,WL_Knowledge_Base_2,WL_Knowledge_Base_1
       INBOUND_QUEUE: RULE_RUNNER_1
-      OUTBOUND_QUEUE: GTAS_RULE_ENGINE
+      OUTBOUND_QUEUE: GTAS_RULE_ENSGINE
```
originalname51, line 112: appears to be misspelled

gtas-parent/gtas-commons/src/main/resources/default.application.properties
```diff
@@ -254,13 +254,15 @@ admin.list.of.integers=1
 #Additional Processing
 additional.processing.enabled.raw=false
 additional.processing.enabled.rules=false
-additional.processing.enabled.passenger=false
+additional.processing.enabled.passenger=true
```
originalname51, line 257: set to false

simbamarufu1:
LGTM

PR #2076, opened 2021-08-20
Refactor logic related to misplaced passwords

simbamarufu1:


originalname51 (approved):
We may want to consider deprecating or updating the fetchUser method in the future.

PR #2077, opened 2021-08-24
Update PVL Page Load

originalname51:
** Update from using a loop to lazily load collections to instead using
a query to instantiate the needed data.
** Removed "transaction" annotation from method as lazy load exceptions
will no longer happen as data is no longer loaded in a loop.
** Update passenger repository with custom query.


simbamarufu1 (commented):
LGTM

simbamarufu1 (approved):

PR #2081, opened 2021-09-16
Update POE query.

originalname51:
MariaDB operator precedence was pulling back more records than needed
for the POE. Update to always have filters.


tjwdev (approved):
LGTM!

